PR #105, opened 2020-01-08
Additional http metrics

aecay:
We want this to be able to add info about expected delays to the http metrics when doing chaos testing.

(Optional) This PR also changes the name of the metrics we send to remove `curl` (and be more backend agnostic).  Not a key part of the change, however.

### DIFF
--- a/mite/__main__.py
+++ b/mite/__main__.py
@@ -71,7 +71,6 @@ import uvloop
 import msgpack
 
 from .cli.common import _create_config_manager
-from .cli.duplicator import duplicator
 from .cli.stats import stats
 from .collector import Collector
 from .controller import Controller
@@ -120,6 +119,12 @@ def _create_runner_transport(opts):
 def _create_controller_server(opts):
     socket = opts['--controller-socket']
     return _msg_backend_module(opts).ControllerServer(socket)
+
+
+def _create_duplicator(opts):
+    return _msg_backend_module(opts).Duplicator(
+        opts['--message-socket'], opts['OUT_SOCKET']
+    )
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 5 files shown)


jb098:
Conceptually I like it because I think one of our key selling points, behind being crazy fast, should be complete felxibility. Should we do this more generally and just implement **additional_metrics into all stats receivers, beyond just the http one?

jb098:
Effectively make it part of the stats 'api'? We don't really have one beyond picking up sent messaes and doing what you like but we could suggest this as a pattern than people might want to use when making their own stats receivers?

I forgot that we could have multiple stats receivers so we can't just have a generic one because it would be picked up by all stats receivers, but I still think making thins more widespread would eb a good idea.

aecay:
This isn't about the stats receiver (yet...I hadn't gotten that far in my thinking yet).  This is just modifying the sending side of the mite_http module.  Currently we send a bunch of predefined info (total time, DNS time, TCP connect time, ...).  This PR allows journeys to add their own custom metrics to what is sent (in this case, expected delay time).

We'll have to create a custom stats receiver (using the machinery from the recent stats changes that we brought in with the AMQP adapter).  But that will sit in id-mite-nft (at least to begin with).  It would make a good case study for the docs once it's fully spec'd out and implemented, though... 🙂

jb098:
I guess you've convinced me. I'd usually be nervous about doing something in mite that has a counterpart in our test repo. I've not looked in great detail but it would be nice to look at a receiver that could just receive and process anything that is sent to it. Seems to me that's something that people would want.

jb098:
This is marked as 'draft', what do you want me to do? Approve here or wait for full PR?

aecay:
I'll unmark it as a draft tomorrow, when the PR for id-mite-nft also goes in.  (I was hoping to get to it this afternoon, but I ran out of time).  I also need to check why the CI failed...

PR #106, opened 2020-01-09
fix tests and implementation of stats.py

aecay:


mite/stats.py
```diff
@@ -97,7 +97,7 @@ def process(self, msg):
                 self.sums[key] += value
                 self.total_counts[key] += 1
                 for i, bin in enumerate(self.bins):
-                    if bin <= value:
```
jb098: Lol

aecay: wasnt-me

PR #107, opened 2020-01-09
Cleanup the cd.yml file

aecay:
- Use our own slave type
- Remove unneeded job

### DIFF
--- a/cd.yml
+++ b/cd.yml
@@ -3,14 +3,9 @@ context: identity
 triggering: multibranch
 
 defaultNodes:
-  default: id-small-slave
+  default: id-mite-slave
 
 modules:
-  mite:
-    pipelineShape: simple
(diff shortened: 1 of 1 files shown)


cd.yml
```diff
@@ -3,14 +3,9 @@ context: identity
 triggering: multibranch
 
 defaultNodes:
-  default: id-small-slave
+  default: id-mite-slave
```
jb098: We going back to our own slave?

aecay: We never used our own slave in this repo, AFAIK.  We got it set up for id-mite-nft, but never used it here.  It works on id-small-slave, so this change isn't completely necessary -- but it makes things easier for us (e.g. if we want to make sure we're always on the latest point release of python for running the unit tests)

PR #108, opened 2020-01-09
Add the ability to dump some messages for debugging to the duplicator

aecay:
This allows us to do `kill -USR1 <pid>` and force the duplicator to dump
100 messages to stdout.  This is useful for debugging live systems,
especially in the cloud where inserting debugger breakpoints isnʼt
feasible.

Probably needs to be performance benchmarked.

### DIFF
--- a/mite/__main__.py
+++ b/mite/__main__.py
@@ -71,6 +71,7 @@ import uvloop
 import msgpack
 
 from .cli.common import _create_config_manager
+from .cli.duplicator import duplicator
 from .cli.stats import stats
 from .collector import Collector
 from .controller import Controller
@@ -119,12 +120,6 @@ def _create_runner_transport(opts):
 def _create_controller_server(opts):
     socket = opts['--controller-socket']
     return _msg_backend_module(opts).ControllerServer(socket)
-
-
-def _create_duplicator(opts):
-    return _msg_backend_module(opts).Duplicator(
-        opts['--message-socket'], opts['OUT_SOCKET']
-    )
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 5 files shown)


jb098:
Is 100 enough for a duplicator churning that much message throughput?

jb098:
Is there a nice way to set it in an env var or something? I suppose you don't want to dummp too much but there's always a good old pipe to file if you want 10000 lines.

aecay:
Great idea to use an env variable.  However, it will have to be set when the duplicator starts, since we can't edit the environment from outside once the process is running.  An alternative would be to have it read the value from a file somewhere, which we can set dynamically... 🤔

jb098:
Have it pull an env variable but default if not there and then run it later?

aecay:
OK, so now the configuration is as follows:
- if the env variable `MITE_DEBUG_MESSAGES_TO_DUMP` exists and is an integer, use that
- else, if the env variable MITE_DEBUG_MESSAGES_TO_DUMP_FILE` exists, try to read an integer value from it and if successful use that
- otherwise, just use 100

I think that covers all the possible use cases...

aecay:
I ran the perftest suite twice for master, and twice for this branch.  The suite is really noisy in general (😿 ), but there's no evidence to suggest that the changes introduced here are negatively impacting the performance seriously enough to be detectable outside the noise.  So, I'm relatively comfortable with these changes from a perf POV.

(Numbers in the table are the total message throughput in 1 minute when sending N messages concurrently from a controller->runner->duplicator->collector.  All the components in this pipeline are run as a single instance.)

| Concurrent messages | master | master2 | message dumper | message dumper2 |
|---------------------|--------|---------|----------------|-----------------|
|                   1 |  51346 |   40580 |          47353 |           51976 |
|                  10 | 183384 |  144722 |         168936 |          168403 |
|                 100 | 360884 |  297529 |         316566 |          316913 |
|                1000 | 333000 |  342346 |         320112 |          312281 |

PR #118, opened 2020-02-10
Fix a problem with mite scenario test

aecay:
This is related to some of the recent async changes.  Iʼm not sure
how/why it hasnʼt turned up til now, but here it is.... *shrug*


jb098:
Not a clue how this didn't break everything.

Issue #114, opened 2020-02-03
Fractional and per-time-unit scheduling


aecay:
Mite's current job scheduling model is based on maintaining a certain number of jobs active at once.  This means that our jobs need to have sleeps in them (boilerplate); it also means that there is no good way to schedule a very-occasional request without a long sleep (which causes problems for the controller teardown).  We also model the work that runners are doing by the number of jobs scheduled on them.  A job that is sleeping doesn't represent load in the same way that a job doing network I/O does (even though both are in an identical "waiting" state, one consumes network resources and the other doesn't).  That means that our distribution of work across runners won't be optimal.

We should also support a job scheduling model that gives an interval over which jobs should run (each second, run 5 of this journey).  The advantage to this model is that it generalizes easily to occasional journeys (every sixty seconds, run one copy of this journey) and one-shot jobs (at 1:00 into the test, run this journey).  I think we can do this, on the controller end, with a token-bucket model: for each journey, we have an asyncio queue.  One coroutine in a forever-loop does { deposit N tokens; sleep M seconds }.  Another blocks on the queue, and whenever it can take tokens out, kicks off the jobs.  The existing model might be adaptable into this framework as well: instead of depositing tokens on a time basis, we'd deposit them back into the bucket when the job completed.

PR #119, opened 2020-03-02
Time function

aecay:
Add a concept of "time function" to mite.  This is a function that is run in the controller alongside the main controller logic.  It receives two `asyncio.Event` objects: one which it can use to gate the start of the test, and one it can use to listen for the end of the test.  We are able to port the previously added "logging webhook" feature to be a generic time function.  We have other use cases in our private testing code as well.  One is at https://github.com/sky-uk/id-mite-nft/blob/8e8f8e10672d0903aceb253d81dd600fe8a65390/mite_id/ims/journeys.py#L200 (where we purge rabbit queues at the start of a test; currently we just use the beginning of the scenario function for this) and for our upcoming chaos testing deliverables (where we want to run not only at the beginning of the test, but during it as well).

Main changes:
- implementation logic for the above
- add a decorator `time_function` to the mite top-level module.  When applied to a scenario function, it marks the scenario to be used with the time function.
- port the logging webhook to be a time function instead (now in `lib/logging_webhook.py`)
- add some unit tests for this functionality (principally, the logic of `asyncio.Event` sequencing)

Consequent changes:
- `mite/__init__.py`: add underscore prefixes to private names
- move the controller logic from `__main__.py` to a new file `cli/controller.py`
- refactor `ScenarioManager` to know the spec of the scenario it represents.  `Controller` objects no longer need to know this; they can read it from their ScenarioManager.
- refactor the `mite controller` and `mite (journey|scenario) test` CLI commands to use the same code path.  This allows changes like this one introducing time functions to be made in only one place, rather than two.

Outstanding tasks:
- [ ] documentation
- [ ] consider whether, for the sake of modularity/composability, we should allow multiple time functions to be specified.
- [ ] consider whether to allow specifying a time function on the command line and/or via environment variables, in addition to as a decorator
- [ ] consider whether there's a better name than "time function" for this feature

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,7 +18,6 @@ dist/
 downloads/
 eggs/
 .eggs/
-lib/
 lib64/
 parts/
 sdist/
--- a/mite/__init__.py
+++ b/mite/__init__.py
@@ -9,6 +9,7 @@ import asyncio
 import random
 from pkg_resources import get_distribution, DistributionNotFound
 
+from .scenario import time_function  # noqa: F401
 from .exceptions import MiteError  # noqa: F401
 from .context import Context
 import mite.utils
@@ -29,7 +30,7 @@ def test_context(extensions=('http',), **config):
     return c
 
 
-class ensure_separation_from_callable:
+class _ensure_separation_from_callable:
     def __init__(self, sep_callable, loop=None):
         self._sep_callable = sep_callable
         self._loop = loop
@@ -65,7 +66,7 @@ def ensure_fixed_separation(separation, loop=None):
     def fixed_separation():
         return separation
 
-    return ensure_separation_from_callable(fixed_separation, loop=loop)
+    return _ensure_separation_from_callable(fixed_separation, loop=loop)
 
 
 def ensure_average_separation(mean_separation, plus_minus=None, loop=None):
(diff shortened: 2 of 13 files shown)


jb098:
One of the problems I see with this are around seperation of concerns. The controller's job should be to control tasks and not run them in my opinion. While it may seem like a bit of a cludge to add something to the beginning of a scenario to do something like cleanup data from an environment, that is still a 'task' which is run according to some sort of schedule.

There is also the concern that the controller is one of the only bottlenecking areas in mite, the more we put into it, the more we risk bringing that bottleneck closer to the inevitable having to deal with it stage.

I'd maybe like to chat about the problem we're trying to solve, perhaps if we're looking to trigger behaviour based on some event listener, we could explore if a change to the the logic in the runners might make a better place for it by swapping the volume model out for something else? I'm not sure whether that's appropriate as it seems you care very much about when a 'test' starts and ends which the controller does understand best.

Let's just think about it a little is all I'm saying :)

jb098:
I also agree with the sentiments around naming 😄

aecay:
Closing this PR for now.  As we discussed, we're going to do this a different way, via changes to the volume model abstraction.  That will allow us to keep tasks on the runners (not the controller) and also bring a fix for #114 into the scope of the resolution as well.

PR #120, opened 2020-03-04
Several changes

aecay:
Substantive changes:
- Change the semantics of the `Gauge` stat to report the last-seen value, not a sum.  We have the "Counter" and "Accumulator" stats for reporting sums.
  - Consequent change to the tests
- Fix a bug whereby the controller was not sending its reports, and therefore the `mite_runner_count` metric (and a couple others) would not appear in prometheus
  - Add a test for this
- Allow the scenario function to receive a sender object for the mite data pipeline if needed.  Dependency-inject this, and the config argument, "properly" (by looking at the signature of the function).

Test changes:
- introduce a "slow" mark for tests.  You can skip the slow tests locally by running `pytest -m "not slow"`
- fix a deprecation warning in the test suite from msgpack library

At https://github.com/sky-uk/mite/pull/120/files#diff-f768c2d8aec809f495425dbf2bdb92f3R88 we have an example of running the "real" runner, duplicator, and controller with the multiprocessing module.  This might be interesting beyond the test suite, so I call attention to it here.

The above changes are to support the chaos testing with tc in our private test suite code.

### DIFF
--- a/acurl/tests/test_httpbin.py
+++ b/acurl/tests/test_httpbin.py
@@ -1,5 +1,6 @@
-import acurl
 from urllib.parse import urlencode
+
+import acurl
 import pytest
 
 
@@ -9,6 +10,7 @@ def session():
 
 
 @pytest.mark.asyncio
+@pytest.mark.slow
 async def test_get():
     s = session()
     r = await s.get('https://httpbin.org/ip')
@@ -19,6 +21,7 @@ async def test_get():
 
 
 @pytest.mark.asyncio
+@pytest.mark.slow
 async def test_cookies():
     s = session()
     r = await s.get('https://httpbin.org/cookies/set?name=value')
@@ -26,6 +29,7 @@ async def test_cookies():
 
 
 @pytest.mark.asyncio
+@pytest.mark.slow
 async def test_session_cookies():
     s = session()
     await s.get('https://httpbin.org/cookies/set?name=value')
@@ -38,6 +42,7 @@ async def test_session_cookies():
 
 
 @pytest.mark.asyncio
+@pytest.mark.slow
 async def test_set_cookies():
     s = session()
     await s.get('https://httpbin.org/cookies/set?name=value')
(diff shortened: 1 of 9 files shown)


jb098:
There's nothing inside our various stats modules that expected the previous behaviour of the Gauge?

aecay:
I don't think so.  The "dump" method of the Gauge class in stats.py just turns the value into a number.  That's ingested by the Gauge class in web/prometheus.py, but that just copies the value over from whatever it receives.

There's a bit of a code smell in terms of the duplication of logic between stats.py and web/prometheus.py, but I think that's because (for better or for worse) we're doing the aggregation in two stages.  Each `mite stats` process receives multiple messages from the runner/controller and aggregates them into one message to `mite exporter`.  The single `mite exporter` receives multiple messages from `mite stats` (either from multiple processes, or multiple times in the same scrape interval) and aggregates them into a single view when the prometheus scrape request comes in.

(My use case for the new kind of gauges is to record the latency introduced by tc.  We don't want to do any kind of summation there; we just have "whatever the latency was last set to on the pods").

aecay:
Substantive changes:
- Change the semantics of the `Gauge` stat to report the last-seen value, not a sum.  We have the "Counter" and "Accumulator" stats for reporting sums.
  - Consequent change to the tests
- Fix a bug whereby the controller was not sending its reports, and therefore the `mite_runner_count` metric (and a couple others) would not appear in prometheus
  - Add a test for this
- Allow the scenario function to receive a sender object for the mite data pipeline if needed.  Dependency-inject this, and the config argument, "properly" (by looking at the signature of the function).

Test changes:
- introduce a "slow" mark for tests.  You can skip the slow tests locally by running `pytest -m "not slow"`
- fix a deprecation warning in the test suite from msgpack library

At https://github.com/sky-uk/mite/pull/120/files#diff-f768c2d8aec809f495425dbf2bdb92f3R88 we have an example of running the "real" runner, duplicator, and controller with the multiprocessing module.  This might be interesting beyond the test suite, so I call attention to it here.

The above changes are to support the chaos testing with tc in our private test suite code.

PR #121, opened 2020-03-10
Misc fixes

aecay:
- add isort to the pre-commit hooks
- run all the pre-commit checks (not only flake8) as part of CI
- rename `test/test_prometheus.py` to avoid a naming conflict.  We need to have unique names of test files across all our repos so we can run the tests in a suite together.  ([The alternative](https://docs.pytest.org/en/latest/goodpractices.html#choosing-a-test-layout-import-rules) of adding `__init__.py` files to the test directories is not palatable).


.isort.cfg
```diff
@@ -0,0 +1,8 @@
+[settings]
+multi_line_output=3
+include_trailing_comma=True
+force_grid_wrap=0
+use_parentheses=True
+line_length=88
+known_first_party=_acurl
+known_third_party = aio_pika,altair,asyncmock,bs4,docopt,flask,ipdb,jinja2,mocks,msgpack,nanomsg,pandas,pkg_resources,psutil,pytest,pytest_httpserver,sanic,selenium,setuptools,ujson,uvloop,werkzeug,zmq
```
jb098: What supreme laziness is this :'D

mite/test.py
```diff
@@ -72,19 +73,20 @@ def http_spy(journey):
...
         return {'http': requests}
+
     return wrapper
 
 
 async def run_single_journey(config, journey, datapool=None):
     messages = []
 
-    async def _send(message, **kwargs):
+    def _send(message, **kwargs):
```
jb098: How was this working?

.isort.cfg
aecay: isort is a package that sorts imports at the top of the file into sections: stdlib, third-party imports, and first-party (i.e. from the same package).  It tries to detect what is third-party and what is first-party, but it doesn't always do a good job at distinguishing.  So we specify in the config file which is which.

Speaking of supreme laziness, I did not write the config manually, but found a tool that does it automatically (otherwise I would not have bothered).  And that's all plumbed in as a pre-commit hook, so it will always stay up to date and I never have to think about it again.  That's the theory at least.  Supreme laziness to the max :P

mite/test.py
aecay: It's probably left over from my attempt to introduce more asynchronicity in the zmq methods, which I had to back out when it didn't exactly work.  This code is to support unit tests external to mite, so it's not been exercised since then until now as I'm trying to resurrect the id-mite-nft unit test suite...

PR #123, opened 2020-03-26
Attempt to fix a race with the runner exit

aecay:
See the comment in the code for explanation

### DIFF
--- a/mite/__main__.py
+++ b/mite/__main__.py
@@ -69,7 +69,6 @@ from urllib.request import urlopen
 import docopt
 import msgpack
 import ujson
-
 import uvloop
 
 from .cli.common import _create_config_manager, _create_runner, _create_scenario_manager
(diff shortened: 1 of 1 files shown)


mite/__main__.py
```diff
@@ -257,9 +256,22 @@ def controller(opts):
...
+    # RuntimeError: Event loop stopped before Future completed.
+    # I think this comes about becasue of a race where the callback scheduled
+    # by RunnerTransport.bye does not get serviced before the above
+    # run_until_complete returns.  I'm mystified as to how this can occur
+    # (because bye awaits the callback, so it should complete....)
+    # Nonetheless, the exception happens, and I believe that this is the
+    # cause.  So, this is an attempt to defend against that error case.  We
+    # give the loop 5 seconds to complete any network ops that are outstanding
+    # before calling the close method which will cancel any scheduled
+    # callbacks and should ensure that the porgrma exits cleanly.
```
jb098: Also, porgrma is not a word. And I just spotted becasue on line 264.

aecay:
I picked 5 because it seems like extremely ample time to complete whatever network I/O might be outstanding, but insignificant from the perspective of our overnight test pipelines.  (For comparison, we give overnight tests a 10 minute grace period to finish after they are "supposed" to be done, so 5s is really a rounding error in those terms)

PR #124, opened 2020-04-01
Exit scenario test when work finished

cosmaprc:
Add functionality to stop scenario test when work finished and return status code based on nr of total errors

### DIFF
--- a/mite/__main__.py
+++ b/mite/__main__.py
@@ -2,7 +2,7 @@
 Mite Load Test Framework.
 
 Usage:
-    mite [options] scenario test [--add-to-config=NEW_VALUE]... [--message-processors=PROCESSORS] [(--should-stop=BOOL | (--should-stop=BOOL --allowed-errors=NUMBER | --should-stop=BOOL --allowed-errors-pct=NUMBER))] SCENARIO_SPEC
+    mite [options] scenario test [--add-to-config=NEW_VALUE]... [--message-processors=PROCESSORS] SCENARIO_SPEC
     mite [options] journey test [--add-to-config=NEW_VALUE]... [--message-processors=PROCESSORS] JOURNEY_SPEC [DATAPOOL_SPEC]
     mite [options] controller SCENARIO_SPEC [--message-socket=SOCKET] [--controller-socket=SOCKET] [--logging-webhook=URL] [--add-to-config=NEW_VALUE]...
     mite [options] runner [--message-socket=SOCKET] [--controller-socket=SOCKET]
@@ -56,9 +56,6 @@ Options:
     --sleep-time=SLEEP                Set the second to await between each request [default: 1]
     --logging-webhook=URL             URL of an HTTP server to log test runs to
     --message-processors=PROCESSORS   Classes to connect to the message bus for local testing [default: mite.logoutput:HttpStatsOutput,mite.logoutput:MsgOutput]
-    --should-stop=BOOL                Stop scenario test if controller should stop. True or False. [default: False]
-    --allowed-errors=NUMBER           On scenarion test stop, if total requests was 0 or more than allowed nr. of errors occurred, exit with status code `1`. NUMBER >= 0. (used with --should-stop, requires mite.logoutput:HttpStatsOutput message processor)
-    --allowed-errors-pct=NUMBER       On scenarion test stop, if total requests was 0 or more than allowed % of total requests were errors, exit with status code `1`. NUMBER between 0 and 100. (used with --should-stop, requires mite.logoutput:HttpStatsOutput message processor)
 """
 import asyncio
 import inspect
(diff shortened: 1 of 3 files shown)


mite/__main__.py
```diff
@@ -56,6 +56,9 @@
     --sleep-time=SLEEP                Set the second to await between each request [default: 1]
     --logging-webhook=URL             URL of an HTTP server to log test runs to
     --message-processors=PROCESSORS   Classes to connect to the message bus for local testing [default: mite.logoutput:HttpStatsOutput,mite.logoutput:MsgOutput]
+    --should-stop=BOOL                Stop scenario test if controller should stop. True or False. [default: False]
```
aecay, line 59: I don't think should_stop should be an option, it should just always happen.

mite/__main__.py
```diff
@@ -2,7 +2,7 @@
 Mite Load Test Framework.
 
 Usage:
-    mite [options] scenario test [--add-to-config=NEW_VALUE]... [--message-processors=PROCESSORS] SCENARIO_SPEC
+    mite [options] scenario test [--add-to-config=NEW_VALUE]... [--message-processors=PROCESSORS] [(--should-stop=BOOL | (--should-stop=BOOL --allowed-errors=NUMBER | --should-stop=BOOL --allowed-errors-pct=NUMBER))] SCENARIO_SPEC
```
aecay, line 5:
```suggestion
    mite [options] scenario test [--add-to-config=NEW_VALUE]... [--message-processors=PROCESSORS] [--allowed-errors=NUMBER] [--allowed-errors-pct=NUMBER] SCENARIO_SPEC
```

mite/__main__.py
```diff
@@ -56,6 +56,9 @@
     --sleep-time=SLEEP                Set the second to await between each request [default: 1]
     --logging-webhook=URL             URL of an HTTP server to log test runs to
     --message-processors=PROCESSORS   Classes to connect to the message bus for local testing [default: mite.logoutput:HttpStatsOutput,mite.logoutput:MsgOutput]
+    --should-stop=BOOL                Stop scenario test if controller should stop. True or False. [default: False]
+    --allowed-errors=NUMBER           On scenarion test stop, if total requests was 0 or more than allowed nr. of errors occurred, exit with status code `1`. NUMBER >= 0. (used with --should-stop, requires mite.logoutput:HttpStatsOutput message processor)
```
aecay, line 60:
```suggestion
    --allowed-errors=NUMBER           On scenario test completion, if total requests was 0 or more than allowed nr. of errors occurred, exit with status code `1`. (requires mite.logoutput:HttpStatsOutput message processor)
```

mite/__main__.py
```diff
@@ -56,6 +56,9 @@
...
+    --allowed-errors-pct=NUMBER       On scenarion test stop, if total requests was 0 or more than allowed % of total requests were errors, exit with status code `1`. NUMBER between 0 and 100. (used with --should-stop, requires mite.logoutput:HttpStatsOutput message processor)
```
aecay, line 61:
```suggestion
    --allowed-errors-pct=NUMBER       On scenario test completion, if total requests was 0 or more than this % of total requests were errors, exit with status code `1`. (requires mite.logoutput:HttpStatsOutput message processor)
```

mite/__main__.py
```diff
@@ -321,10 +324,29 @@ def har_converter(opts):
     )
 
 
+def validate_opts(opts):
+
+    try:
+        assert opts['--should-stop'] in ['True', 'False']
+    except Exception:
```
aecay, line 331:
```suggestion
    if opts['--should-stop'] not in ['True', 'False']:
```

mite/__main__.py
```diff
@@ -321,10 +324,29 @@ def har_converter(opts):
...
+        raise docopt.DocoptExit("--should-stop has to be True or False")
+
+    try:
+        assert (lambda e: e is None or int(e) >= 0)(opts['--allowed-errors'])
+    except Exception:
```
aecay, line 336:
```suggestion
    if opts['--allowed-errors'] is None or not opts['--allowed-errors'].isdigit()
```

mite/__main__.py
```diff
@@ -321,10 +324,29 @@ def har_converter(opts):
...
+        raise docopt.DocoptExit("--allowed-errors has to be >= 0")
+
+    try:
+        assert (lambda e: e is None or 0 <= float(e) <= 100)(opts['--allowed-errors-pct'])
```
aecay, line 340:
```suggestion
        if opts['--allowed-errors-pct'] is not None:
            pct = float(opts['--allowed-errors-pct'])
            assert 0 <= pct <= 100
```

cosmaprc:
> Left some comments on the code. But I wonder whether this whole thing is getting too complicated. For your purposes for the data creation, does it suffice for the command to exit `1` if there were any errors, and `0` if there were none?
> 
> If that is sufficient, let's just implement that, rather than adding extra command line options etc.

Agreed, have pushed 

> Left some comments on the code. But I wonder whether this whole thing is getting too complicated. For your purposes for the data creation, does it suffice for the command to exit `1` if there were any errors, and `0` if there were none?
> 
> If that is sufficient, let's just implement that, rather than adding extra command line options etc.

Agreed. I have pushed review fixes removing the new cmd args and just changed the func of the scenario test to exit when it's done and return 1 if there are errors

mite/cli/test.py
```diff
@@ -50,17 +59,22 @@ def _setup_msg_processors(receiver, opts):
...
         else:
-            logging.error(f"Class {processor.__name__} does not have a process(_raw)_message method!")
+            logging.error(
+                f"Class {processor.__name__} does not have a process(_raw)_message method!"
+            )
+
+
+def _get_http_stats_output(receiver):
+    listeners = receiver.filter_listeners(HttpStatsOutput)
+    assert len(listeners) == 1
```
aecay, line 76: Oh, one more that I forgot the first time (sorry) -- this will fail if `HttpStatsOutput` isn;t one fo the specified processors.  I think we should return `None` from this function in that case -- and skip the error validation (i.e. always just `return` from the scenario test command)

mite/cli/test.py
```diff
@@ -50,17 +59,21 @@ def _setup_msg_processors(receiver, opts):
...
         elif hasattr(processor, "process_raw_message"):
             receiver.add_raw_listener(processor.process_raw_message)
         else:
-            logging.error(f"Class {processor.__name__} does not have a process(_raw)_message method!")
+            logging.error(
+                f"Class {processor.__name__} does not have a process(_raw)_message method!"
+            )
+
+
+def _get_http_stats_output(receiver):
```
jb098, line 74: So this function is used to return a listener if there is one and None if not? Is there any case where there's more than one listener?

PR #125, opened 2020-04-21
Allow colons in config values

jarnfield:


aecay:
The CI failure is because we pinned to an older version of pytest for compatibility with python 3.8 betas.  Now 3.8 is out officially, so I think we can remove that pin now (which broke because pytest-asyncio recently upgraded to an API not available in our chosen pytest version).

I think you can just remove the version requirement on pytest in `test-requirements.txt` and repush.  If the tests pass then it's good to go 🤞

PR #126, opened 2020-04-27
First commit for a mite websocket adaptor

jb098:
* Pretty lightweight and similar to the amqp one
* Adds websocket attribute to conext using wrapper
* Uses connect method to establish websocket connection, which both returns the connection object and sets a websocket.connection attribute to the established connection
* websocket.send(...) will then send over this connection
* if more connections are needed, you can assign returned connection from connect(...). Unsure if this is best design
* 100% unit test coverage
* Cool new exception chaining feature I've never used before

### DIFF
--- a/mite_websocket/__init__.py
+++ b/mite_websocket/__init__.py
@@ -1,0 +1,53 @@
+import asyncio
+import logging
+from contextlib import asynccontextmanager
+
+import websockets
+from websockets.exceptions import WebSocketException
+
+from mite.exceptions import MiteError
+
+
+logger = logging.getLogger(__name__) 
+
+
+class WebsocketError(MiteError): 
+    pass
+
+
+class _WebsocketWrapper:
+    def __init__(self):
+        self.connection = None
+
+    def install(self, context):
+        context.websocket = self
+
+    def uninstall(self, context):
+        del context.websocket
+
+    async def connect(self, *args, **kwargs):
+        self.connection = await websockets.connect(*args, **kwargs)
+        return self.connection 
+
+    async def send(self, body, **kwargs):
+        await self.connection.send(body)
+
+
+@asynccontextmanager
+async def _websocket_context_manager(context):
+    ww = _WebsocketWrapper()
+    ww.install(context)
+    try:
+        yield
+    except WebSocketException as e:
+        raise WebsocketError(e) from e
+    finally:
+        ww.uninstall(context)
+
+
+def mite_websocket(func):
+    async def wrapper(ctx, *args, **kwargs):
+        async with _websocket_context_manager(ctx):
+            return await func(ctx, *args, **kwargs)
+
+    return wrapper
--- a/test/test_mite_websockets.py
+++ b/test/test_mite_websockets.py
@@ -1,0 +1,67 @@
+import asyncio
+from unittest.mock import patch
+
+import pytest
+from asyncmock import AsyncMock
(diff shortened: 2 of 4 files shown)


aecay:
CI is grumpy though...  Looks like small things, unused imports and the like

jb098:
> CI is grumpy though... Looks like small things, unused imports and the like

Stylistic fixes have been made.

test/test_mite_websockets.py
```diff
@@ -0,0 +1,63 @@
...
+
+    assert getattr(context, "websocket", None) is None
+
+
+@pytest.mark.asyncio
+async def test_mite_websocket_connect_and_send():
+    context = MockContext()
+    url = "wss://foo.bar"
+    message = "baz"
+    connect_mock = AsyncMock()
```
haffi96, line 40: What is this connect_mock argument testing for?

test/test_mite_websockets.py
```diff
@@ -0,0 +1,63 @@
...
+
+    @mite_websocket
+    async def dummy_journey(ctx):
+        await ctx.websocket.connect(url)
+        await ctx.websocket.send(message)
+
+    with patch("websockets.connect", new=connect_mock):
```
haffi96, line 47: Can you explain what is this testing?

mite_websocket/__init__.py
```diff
@@ -0,0 +1,52 @@
...
+@asynccontextmanager
+async def _websocket_context_manager(context):
+    ww = _WebsocketWrapper()
+    ww.install(context)
+    try:
+        yield
+    except WebSocketException as e:
+        raise WebsocketError(e) from e
+    finally:
+        ww.uninstall(context)
```
haffi96, line 44: Is it just good practice to del connection after?

mite_websocket/__init__.py
```diff
@@ -0,0 +1,52 @@
...
+        self.connection = None
+
+    def install(self, context):
+        context.websocket = self
+
+    def uninstall(self, context):
+        del context.websocket
+
+    async def connect(self, *args, **kwargs):
+        self.connection = await websockets.connect(*args, **kwargs)
```
jarnfield, line 28: When do we disconnect? Is it worth using the connection's context manager as documented on the websocket home page to connect and disconnect automatically when entering and exiting a relevant journey?

jb098: It's a good point @jarnfield, I'll add something to do this

test/test_mite_websockets.py
jb098: Sure, so the mocking stuff in Python is pretty powerful. What I'm saying here is that if at any point in this block, a call is made to websockets.connect, don't run that function, return a mock instead. My code will run and be tested as normal, but the underlying call to create the connection will never actually be made. 

The mocks returned by default can be called with any method and they'll be monkey patched on. So in this case once ctx.websocket.connection is set to a mock, I call my send method which calls the send method on the 'connection' which is actually a mock and itself returns another mock. Then I can check that both the connect and the send method were called as I expected. 

I then have to have some degree of faith the the websockets library works as desribed and that they've done their own testing.

mite_websocket/__init__.py
jb098: I'll disconnect it, don't think we need to actualy delete it. It will be collected when the parent object is no longer referenced.

PR #127, opened 2020-04-30
Fix the url regex in mite browser

aecay:
This has been broken for a long time, but itʼs only now started to
actually cause exceptions...Iʼm not 100% sure why, but letʼs just fix
it.

### DIFF
--- a/mite_browser/__init__.py
+++ b/mite_browser/__init__.py
@@ -1,14 +1,16 @@
 import asyncio
-from bs4 import BeautifulSoup
+import re
 from urllib.parse import urlencode, urljoin
-from re import compile as re_compile, IGNORECASE
+
+from bs4 import BeautifulSoup
+
+import mite_http
 from mite import ensure_fixed_separation
 from mite.exceptions import MiteError
-import mite_http
 
-# TODO: fix
-EMBEDDED_URL_REGEX = re_compile(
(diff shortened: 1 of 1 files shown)


mite_browser/__init__.py
```diff
@@ -1,14 +1,16 @@
...
+import mite_http
 from mite import ensure_fixed_separation
 from mite.exceptions import MiteError
-import mite_http
 
-# TODO: fix
-EMBEDDED_URL_REGEX = re_compile(
-    "\(\s*[\\]?[\"']([^\"':.]*:)?([^\"':.]*\.[^\"':.]*)[\\]?[\"']\s*\)", IGNORECASE  # noqa: W605
+EMBEDDED_URL_REGEX = re.compile(
+    "\\(\\s*[\\]?[\"']([^\"':.]*:)?([^\"':.]*\\.[^\"':.]*)[\\]?[\"']\\s*\\)",
```
jarnfield, line 12: I guess an r string might help make this slighly more readable? I think this is the same:
`r"""\(\s*[\]?["']([^"':.]*:)?([^"':.]*\.[^"':.]*)[\]?["']\s*\)"""`

aecay: Good spot, made that change

PR #128, opened 2020-05-15
Fixed bugs in browser, added unit tests

jb098:
* Fixed at least one bug with parsing of elements with a disabled attribute
* Added 30% coverage of unit tests to the mite_browser
* Increased overall unit test coverage by ~4%

### DIFF
--- a/mite_browser/__init__.py
+++ b/mite_browser/__init__.py
@@ -1,5 +1,6 @@
 import asyncio
 import re
+from contextlib import asynccontextmanager
 from urllib.parse import urlencode, urljoin
 
 from bs4 import BeautifulSoup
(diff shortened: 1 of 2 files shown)


mite_browser/__init__.py
```diff
@@ -409,8 +409,11 @@ def __repr__(self):
 
 
 def _field_is_disabled(element):
-    status = element.get('disabled')
-    return status is not None and status.lower() in ['disabled', 'true']
+    status = element.attrs.get('disabled')
+    if status and status.lower() in ['disabled', 'true']:
```
jarnfield, line 413: Bit of a nit. Not sure if you have a preference but I find this is a trap people fall in to without even noticing. Maybe just 

```
`return status and status.lower() in ['disabled', 'true']
```

PR #129, opened 2020-05-27
Mite websocket update

DavAnnunz:
After founding out that in the current state I wasn't able to receive the message from the server I did some changes.

- When a connection is created it will be added on a list and returned. 
- To send and receive messages is enough to use the returned obj from the connection method.
- There is a new method to get the list of created connection
- A new close method as been added. It will close the connection, remove it from the list and delete it. 
- At the end the whole list will be scanned and all the connection closed.

### DIFF
--- a/mite_websocket/__init__.py
+++ b/mite_websocket/__init__.py
@@ -15,25 +15,29 @@ class WebsocketError(MiteError):
 
 class _WebsocketWrapper:
     def __init__(self):
-        self.connection = None
+        self.connections = []
 
     def install(self, context):
         context.websocket = self
 
     async def uninstall(self, context):
-        if self.connection:
-            await self.connection.close()
+        for conn in self.connections:
+            if conn:
+                await self.close_connection(conn)
         del context.websocket
 
     async def connect(self, *args, **kwargs):
-        self.connection = await websockets.connect(*args, **kwargs)
-        return self.connection
+        conn = await websockets.connect(*args, **kwargs)
+        self.connections.append(conn)
+        return conn
 
-    async def send(self, body, **kwargs):
-        await self.connection.send(body)
+    def get_connections(self):
+        return self.connections
 
-    async def recv(self):
-        return await self.connection.recv()
+    async def close_connection(self, conn):
+        await conn.close()
+        self.connections.remove(conn)
+        del conn
 
 
 @asynccontextmanager
--- a/mite_websocket/__init__.py
+++ b/mite_websocket/__init__.py
@@ -32,6 +32,9 @@ class _WebsocketWrapper:
     async def send(self, body, **kwargs):
         await self.connection.send(body)
 
+    async def recv(self):
+        return await self.connection.recv()
+
 
 @asynccontextmanager
 async def _websocket_context_manager(context):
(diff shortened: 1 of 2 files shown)


mite_websocket/__init__.py
```diff
@@ -15,22 +15,29 @@ class WebsocketError(MiteError):
...
 
     def install(self, context):
         context.websocket = self
 
     async def uninstall(self, context):
-        if self.connection:
-            await self.connection.close()
+        for conn in self.connections:
+            if conn:
+                await self.close_connection(conn)
```
aecay, line 26: Is it safe to call close on a connection that is already closed?  Or does the `if conn` test for closedness (which would be slightly unusual, but maybe that's how the websocket library works...?)

mite_websocket/__init__.py
```diff
@@ -15,22 +15,29 @@ class WebsocketError(MiteError):
...
 
-    async def send(self, body, **kwargs):
-        await self.connection.send(body)
+    def get_connections(self):
+        return self.connections
+
+    async def close_connection(self, conn):
+        await conn.close()
+        self.connections.remove(conn)
+        del conn
```
aecay, line 40: `del conn` isn't necessary....

test/test_mite_websockets.py
```diff
@@ -29,28 +29,25 @@ async def dummy_journey(ctx):
         await ctx.websocket.connect("wss://foo.bar")
 
     with patch("websockets.connect", new=connect_mock):
-        await dummy_journey(context)
+        wb = await dummy_journey(context)
```
aecay, line 32: Why are we capturing the return value of the function?  (Which will always be `None`)

test/test_mite_websockets.py
```diff
@@ -29,28 +29,25 @@ async def dummy_journey(ctx):
...
 
     assert getattr(context, "websocket", None) is None
 
 
 @pytest.mark.asyncio
-async def test_mite_websocket_connect_and_send():
```
aecay, line 38: Should we add a test that validates the sending behavior?  Something like
```
def my_test():
    # set up mock / patch / etc.
    ws = await ctx.websocket.connect("ws://example.com")
    connect_mock.assert_called_once_with("ws://example.com")
    await ws.send("ohai")
    connect_mock.return_value.send.assert_called_once_with("ohai")
```

mite_websocket/__init__.py
DavAnnunz: I tried to close the same connection multiple times and no error are showed. There are errors if I try to send or recv on a closed connection but not error is showed to me closing multiple times.

test/test_mite_websockets.py
DavAnnunz: I can try to do something for it but the send and recv functions are from the websockets library. 
Would it be meaningful if I call the send or request and then assert what I decided to return?

aecay: If the connect function is a mock, then its return value won't be a real websocket, but a mock.  That's the brilliance of the python mock library, you replace one thing with a mock and it cascades through the whole logic, making it easy to write tests. 🙂

mite_websocket/__init__.py
aecay: Thanks for checking, this code is OK then 🙂

PR #131, opened 2020-06-16
Add logging to runner cleanup

jarnfield:


aecay:
@jarnfield is this PR ok to close now?

PR #137, opened 2020-07-29
dns stats mite

VMawson:
Creation of DNS stats. This was made to be completely separate from the http stats. 
There are 2 approaches- the one listed here, which involves the use of the `@mite_dns` wrapper (see the example of its use in the `oogway/init.py` in the DNS_Stats_id-mite branch- the PR for the id-mite-nft branch).
Or the removal of lines 30-38 in the mite_dns/init.py file here, and simply using the `mite_dns = mite_dns:_MITE_STATS`
Both approaches have been tested, so dependant upon which is most preferable. 
id-mite-nft branch----
https://github.com/sky-uk/id-mite-nft/pull/942

example on DNS stats on prometheus for a quick test----
https://identity-prometheus-external.dev.ce.us-west-2-aws.npskooniedc.com/graph?g0.range_input=30m&g0.end_input=2020-07-24%2017%3A24&g0.expr=mite_dns_time_sum&g0.tab=0


mite_dns/__init__.py
```diff
@@ -0,0 +1,38 @@
...
+            extractor=extractor(['transaction'], 'dns_time'),
+            bins=bins,
+        ),
+    )
+
+
+_MITE_STATS = _generate_dns_stats()
+
+@asynccontextmanager
+async def _session_pool_context_manager(session_pool, context):
```
cosmaprc, line 31: is this overriding the _session_pool_context_manager fn in _mite_http.__init__.py ? Does it need this here ?

mite_dns/__init__.py
```diff
@@ -0,0 +1,38 @@
...
+    context.http = await session_pool._checkout(context)
+    yield
+    await session_pool._checkin(context.http)
+    del context.http
+
+def mite_dns(func):
+    return SessionPool.decorator(func)
```
aecay, line 38: I don't think this needs to be a decorator.  It reuses a metric that's already present on the http message.  So we can just add the stat; we don't need this.

setup.cfg
```diff
@@ -26,6 +26,7 @@ mite_stats =
     mite = mite.stats:_MITE_STATS
     mite_http = mite_http:_MITE_STATS
     mite_selenium = mite_selenium:_MITE_STATS
+    mite_dns = mite_dns:_MITE_STATS
```
aecay, line 29: I think it makes sense not to add this by default, to keep the default stats as lean as possible.  Other packages (like id-mite-nft) can add this line if they need dns timings.  (I would also not create the new mite_dns module, which seems like it should allow us to operate with the dns protocol directly.  I'd put the code in mite_http so that we have `mite_http:_DNS_STATS`)

VMawson: just to clarify: are you suggesting removal of the mite_dns/init.py file and moving everything over to mite_http? or just removal of this from setup?

tox.ini
```diff
@@ -25,6 +25,7 @@ source =
     mite_selenium
     mite_amqp
     mite_websocket
+    mite_dns
```
aecay, line 28: don't need this

PR #142, opened 2020-08-10
Enhancements and fixes for "mite cat"

aecay:
- Move the cat command to its own file in the mite.cli package
- Fix a bug introduced by the recent msgpack upgrade (`encoding` arg was removed)
- Make the output format of `mite cat` be JSON
- Add an `uncat` command which performs the inverse of the cat operation


aecay:
@jb098 I think there's a (obscure) unix utility called tac that reverses the lines that are piped through it....I could have recycled that pun I suppose, but i doubt that would have been any better tbh... 😝

PR #145, opened 2020-09-03
Log debugging messages from mite and add one time journey run command

jarnfield:


acurl/acurl/__init__.py
```diff
@@ -252,6 +252,8 @@ def to_curl(self):
             + shlex.quote(self.url)
         )
 
+    as_curl = to_curl
```
aecay, line 255: Why?

mite/__main__.py
```diff
@@ -48,6 +49,7 @@
     --recorder-socket=SOCKET          Socket [default: tcp://127.0.0.1:14306]
     --delay-start-seconds=DELAY       Delay start allowing others to connect [default: 0]
     --volume=VOLUME                   Volume to run journey at [default: 1]
+    --duration=SECONDS                Duration to run journey for [default: 60]
```
aecay, line 52: IIUC, this changes the interface of `mite journey test` to stop after 60 seconds by default...I'm skeptical about making this change to the interface.

mite/context.py
```diff
@@ -73,6 +94,8 @@ def send(self, type, **msg):
 
     @asynccontextmanager
     async def transaction(self, name):
+        path = active_transaction.get() + (name, )
+        active_transaction.set(path)
```
aecay, line 98: We're paying the price for a ContextVariable on every transaction?  😕

mite/context.py
```diff
@@ -99,9 +122,17 @@ async def transaction(self, name):
...
                 raise
         finally:
-            self.send('txn', start_time=start_time, end_time=time.time(), had_error=error)
+            self.send(
+                'txn',
+                start_time=start_time,
+                end_time=time.time(),
+                had_error=error,
+                context=path,
+                meta=get_transaction_meta()
```
aecay, line 131: And also paying for extra data in the pipeline for every txn (even if it's just an empty dict by default)? 😕

mite/cli/test.py
```diff
@@ -83,17 +86,22 @@ def test_scenarios(test_name, opts, scenarios, config_manager):
     controller = Controller(test_name, scenario_manager, config_manager)
     transport = DirectRunnerTransport(controller)
     receiver = DirectReciever()
+    meta_processor = TransactionMetadataCollector()
+    receiver.add_listener(meta_processor.process_message)
```
aecay, line 90: This idea of adding a receiver is fundamentally ok, but why don't we just add a receiver that listens for a new message type (`curl_command`, or if you want to be generic something like `console_debug_message`) and prints out those messages as it sees them?  We don't need to accumulate metadata in a contextvar on the transaction etc etc.

aecay (commented):
I don't really like this PR.  I think the design makes us pay a lot of efficiency taxes in the general case for what's fundamentally a debugging purpose.  And I'm not convinced that the generality of the transaction_meta abstraction has enough concrete usecases to make it worthwhile.

I do think there's the germ of the solution in here though, which is just to add a listener to the data pipeline in the command line cases that will print things out, and to drive it from our test code accordingly.

mite/volume_model.py
```diff
@@ -137,3 +137,19 @@ def _volume(self, start, end):
             return c[1](start - c[0], end - c[0])
         except StopIteration:  # pragma: no cover
             raise Exception("should never happen!")
+
+
+def oneshot_vm(when=-1, stop_scenario=False):
+    """A volume model that returns 1 after ``when`` seconds, and zero before."""
```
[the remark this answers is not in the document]
jarnfield: It's not just the docstring I've changed here, but the implementation as well. I've added an extra condition that makes sure it only return 1 once. In my head this was enough to make it a true one shot vm. But was obviously concerned that I may be wrong due to some case I haven't considered. That's why I removed the comment.

acurl/acurl/__init__.py
jarnfield: This was only a quality of life alias for my own selfish reasons. The handful of times I've been in the --debugging console and tried to get a curl I've typed as_curl first. Happy to remove and spend the extra 2 seconds I spend every time I do this correcting myself (or learn faster).

mite/volume_model.py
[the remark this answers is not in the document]
aecay: Ah, I see now that the implementation is different.  

The number returned by the vm corresponds to simultaneous transactions.  I don't know off the top of my head when the VM gets called again (and I think it partially depends on how frequently runners check in, i.e. how long journeys take -- it's not a static constant).  It's possible that the controller will call the VM, see that the answer is `1`, and then spend the next (say) 30 seconds handing out copies of the job to runners without ever calling the VM again to see it has gone back to zero.

I don't know for sure whether that would happen or not, but the design of the system doesn't rule it out.  The only way to get guaranteed oneshot behavior (as currently architected) is to use a datapool.

mite/__main__.py
jarnfield: I can delete if we want, but struggled to see if a journey would ever stop as we had it before, if it doesn't have an exhaustive data pool (does it?). I figured this was probably better than a never-ending test for all of these journeys. But it's likely another lack of understanding thing.

aecay: My use case for this is often to leave it running indefinitely while I look at logs or something in another console, or wait for an error to hit so I can drop into the debugger, or something.  I just hit ctrl-c when I want it to stop.

jarnfield: Removed

mite/cli/test.py
```diff
@@ -83,17 +85,22 @@ def test_scenarios(test_name, opts, scenarios, config_manager):
...
+    debug_message_output = DebugMessageOutput()
+    receiver.add_listener(debug_message_output.process_message)
     _setup_msg_processors(receiver, opts)
     http_stats_output = _get_http_stats_output(receiver)
     loop = asyncio.get_event_loop()
 
+    has_error = False
+
     async def controller_report():
+        nonlocal has_error
```
aecay, line 97: Rather than this `nonlocal` business (which I find a bit confusing, although in the end I guess it's pretty simple), I think we could do things slightly differently...

1. make `controller_report` return its status directly
2. use `asyncio.create_task` to wrap it in a task (which we save a reference to in a variable), rather than passing `controller_report()` directly to `asyncio.gather`
3. use `task_var.result()` to get the result back

I'm not 100% sure this works (i.e. I haven't tested), but I think it will.  It's also a completely optional suggestion: I've written code just like this with `nonlocal` etc before (and vaguely wished for a better way to write it), but this other pattern didn't occur to me until just now.  Totally up to you whether you want to leave it as it is (which is fundamentally fine) or make further changes.  You might not even agree with me that my notional change is any better!

aecay (approved):

aecay:
LGTM with one optional suggestion/comment

cosmaprc (approved):

PR #147, opened 2020-09-14
Add receiver component to docs

cosmaprc:
Add receiver to docs/components

Add receiver to docs/design-deployment

### DIFF
--- a/docs/components.rst
+++ b/docs/components.rst
@@ -100,17 +100,19 @@ dispatch incoming messages to them.
 
 .. code-block:: sh
 
-   POD_NAME=1 mite receiver tcp://127.0.0.1:14303 \
+   mite receiver tcp://127.0.0.1:14303 \
     --processor=my.custom.processors:StatsProcessor \
     --processor=my.custom.processors:CollectorProcessor
 
-   mite receiver tcp://127.0.0.1:14310 --processor=my.custom.processors:PrintProcessor
+   mite receiver tcp://127.0.0.1:14310 \
+    --processor=my.custom.processors:PrintProcessor
 
 **Example custom processors:**
 
 .. code-block:: python
 
     import os
+    from uuid import uuid4
 
     from mite import collector
     from mite.cli import stats
@@ -128,7 +130,7 @@ dispatch incoming messages to them.
 
     class CollectorProcessor:
         def __init__(self):
-            self.collector = collector.Collector(collector_id=os.environ["POD_NAME"])
+            self.collector = collector.Collector(collector_id=str(uuid4()))
 
         def process_raw_message(self, message):
             return self.collector.process_raw_message(message)
--- a/docs/components.rst
+++ b/docs/components.rst
@@ -86,6 +86,57 @@ socket for the duplicator.
 
 .. _msgpack: https://msgpack.org/index.html
 
+.. _receiver-component:
+
+Receiver
+--------
+
+The mite receiver component is a generic mechanism that can perform tasks
+not limited to fulfilling a single role (eg. stats, collector). This allows
+changing a mite pipeline more easily, without introducing new processes and
+network components. A ``mite.cli.receiver`` instance can have multiple custom
+``processors`` connected to it as either listeners or raw listeners and can
+dispatch incoming messages to them.
+
+.. code-block:: sh
+
+   POD_NAME=1 mite receiver tcp://127.0.0.1:14303 \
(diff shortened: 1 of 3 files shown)


docs/components.rst
```diff
@@ -86,6 +86,57 @@ socket for the duplicator.
...
+The mite receiver component is a generic mechanism that can perform tasks
+not limited to fulfilling a single role (eg. stats, collector). This allows
+changing a mite pipeline more easily, without introducing new processes and
+network components. A ``mite.cli.receiver`` instance can have multiple custom
+``processors`` connected to it as either listeners or raw listeners and can
+dispatch incoming messages to them.
+
+.. code-block:: sh
+
+   POD_NAME=1 mite receiver tcp://127.0.0.1:14303 \
```
aecay, line 103: Why is POD_NAME needed?

docs/design-deployment.rst
```diff
@@ -25,6 +25,7 @@ The components that make up mite are:
 - recorder*: logs messages appearing on the mite message bus to a file
 - collector: listens for special messages on the bus and records them
   to a file.  This is used for :ref:`data creation scenarios <data-creation-scenarios>`.
+- receiver*: dispatches incoming messages to connected ``processors``
```
aecay, line 28: Doesn't need an asterisk bc it's not a singleton

docs/design-deployment.rst
```diff
@@ -25,6 +25,7 @@ The components that make up mite are:
 - recorder*: logs messages appearing on the mite message bus to a file
 - collector: listens for special messages on the bus and records them
```
aecay, line 26: While you are here, can you swap collector and recorder?  This was a mistake from the original docs that I just noticed...

docs/design-deployment.rst
```diff
@@ -44,6 +45,7 @@ between the components are represented in the following diagram:
    duplicator -> collector;
    duplicator -> recorder;
    stats -> exporter;
+   duplicator -> receiver1 -> receiver2;
```
aecay, line 48:
```suggestion
   duplicator -> receiver;
```

I'm not totally sure whether we want to add this here, since it's an alternative to the architecture expressed in the diagram...maybe we want to have two diagrams, or draw a box labeled "receiver" around both stats and collector.

docs/design-deployment.rst
```diff
@@ -75,7 +77,7 @@ focus has been on the performance of the applications which we test.
 While not wishing to be profligate, we believe it would be a mistake to
 shrink the injection infrastructure close to its performance limits.
 When resource constraints appear in an NFT exercise, they should stem
-from the system under test and not the test apparatus.  Anything other
+
```
aecay, line 80: accident?

aecay (commented):
these new docs look really great.  just a few sugestions...

docs/components.rst
cosmaprc: it's tied to the CollectorProcessor example class beneath which does `self.collector = collector.Collector(collector_id=os.environ["POD_NAME"])` and I wanted the example to be complete so I put it in... should I take it out ?

docs/design-deployment.rst
cosmaprc: I'll try the box labeled receiver

docs/components.rst
aecay: Yeah, I think it's confusing.  Can the collector id just be a random uuid?  (Or a random unsigned integer, if it is constrained to being an integer?)  @johnarnfield would that work?

jarnfield: It would 👍

docs/design-deployment.rst
cosmaprc: fixed it

docs/components.rst
cosmaprc: I changed the example code to use a uuid

docs/design-deployment.rst
cosmaprc: fixed and rearranged the components by layers, top to bottom and alphabetically per layer

docs/design-deployment.rst
aecay: gh doesn't think so...

aecay (approved):
lgtm, there's still a deleted line according to the github diff but should be an easy fix

docs/design-deployment.rst
cosmaprc: now really fixed

jarnfield (approved):

PR #148, opened 2020-09-15
Adding a session wrapper for additional metrics

DavAnnunz:
Adding a Session wrapper in mite_http in order to allow the session modification without break the original mite session.

### DIFF
--- a/mite_browser/__init__.py
+++ b/mite_browser/__init__.py
@@ -88,7 +88,7 @@ class Browser:
         """Perform a request and return a page object"""
         # Wrap everything in page object
         embedded_res = kwargs.pop("embedded_res", self._embedded_res)
-        resp = await self._session.getattr("request")(method, url, *args, **kwargs)
+        resp = await self._session.request(method, url, *args, **kwargs)
         page = Page(resp, self)
         if embedded_res:
             async with self._ctx.transaction(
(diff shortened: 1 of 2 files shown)


mite_http/__init__.py
```diff
@@ -8,12 +8,27 @@
...
-    del context.http
+class AcurlSessionWrapper:
+    def __init__(self, session):
+        self.__session = session
+        self.__callback = None
+        self.additional_metrics = {}
+
+    def getattr(self, attrname):
+        try:
+            r = object.___getattr__(self, attrname)
```
aecay, line 19:
```suggestion
            r = object.__getattr__(self, attrname)
```
Is this an eye test? 😝

aecay (approved):
lgtm, with one small fix

cosmaprc (approved):

mite_browser/__init__.py
```diff
@@ -88,7 +88,7 @@ async def request(self, method, url, *args, **kwargs):
         """Perform a request and return a page object"""
         # Wrap everything in page object
         embedded_res = kwargs.pop("embedded_res", self._embedded_res)
-        resp = await self._session.request(method, url, *args, **kwargs)
+        resp = await self._session.getattr("request")(method, url, *args, **kwargs)
```
aecay, line 91: Like in my comment on the other PR, if this is necessary then it seems like the wrapper is not quite constructed right...

PR #149, opened 2020-09-15
Add a couple dummy sleep tasks to hack around the controller hang

aecay:
If I'm right in my analysis at https://github.com/sky-uk/id-mite-nft/issues/894#issuecomment-654823633, then having a couple extra timers attached to the event loop should prevent the error condition from occurring.

### DIFF
--- a/mite/__main__.py
+++ b/mite/__main__.py
@@ -199,6 +199,11 @@ def _controller_log_end(logging_id, logging_url):
     logger.debug("Logging test end complete")
 
 
+async def _sleep_loop(delay):
+    while True:
+        await asyncio.sleep(delay)
+
+
 def controller(opts):
     config_manager = _create_config_manager(opts)
     scenario_spec = opts['SCENARIO_SPEC']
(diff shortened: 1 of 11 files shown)


mite/__main__.py
```diff
@@ -236,17 +241,30 @@ def controller(opts):
...
             await asyncio.sleep(1)
             controller.report(sender.send)
 
+    tasks = tuple(
+        asyncio.create_task(x)
+        for x in (
+            # 331 and 337 are prime (and a fortiori relatively prime), so they
+            # should coincide in expiring at the same time only once every ~30
+            # hours (modulo rescheduling jitter/drift...).  This is to prevent
+            # the bug where the controller sometimes fails to exit cleanly.
```
jarnfield, line 253: 😲

jarnfield (approved):
Nice! Fingers crossed.

jb098 (approved):
This is completely mad but I trust your analysis and would love to hear more about it at team showcase.

PR #160, opened 2020-10-28
Remove root transaction

aecay:
With the new nested txn names, we're getting `__root__ ::` prepended to each transaction.  This is one way of solving that.  It changes the semantics of what transactions are, but none of our code relies on there being an outer wrapper transaction (or at least it should not...).  If a user of mite wants a wrapper transaction, that seems like something they can implement themselves through a mechanism similar to our `@legacy_journey` wrapper.

I'm still unsure of whether this is the right solution though...the other alternative would be to add some special-case logic to context.py so that we don't nest the txn name if the parent txn is `__root__`.  Thoughts welcome...

PS I've cheekily snuck an unrelated fix onto this branch in context.py, just cleaning up some no-longer-used variables.  But the change discussed above is in runner.py.

### DIFF
--- a/mite/context.py
+++ b/mite/context.py
@@ -4,22 +4,10 @@ import time
 import traceback
 from contextlib import asynccontextmanager
 from itertools import count
-from pathlib import PurePath
 
 from .exceptions import MiteError
 
 logger = logging.getLogger(__name__)
-
-_HERE = PurePath(__file__).parent
-_MITE_HTTP = _HERE.parent / "mite_http"
-# FIXME: if the mite_http module was installed at mite.http, then we could
-# omit this hack, and just use the (single) directory of the mite library for
-# the checks.  But that would require a change in the way we package mite,
-# which we might or might not ultimately want.  See also:
-# https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies
-# Further FIXME: add the other built-in mite modules to the list of paths to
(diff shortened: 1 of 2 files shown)


jarnfield:
The thing that springs to mind (and this suddenly sounds quite over-engineered) is that there's no need for `self._transaction_name` to be maintained as a string until the time comes to send() where we need to stringify it.

Maybe we can maintain `self._transaction_name` as a stack/list/tuple and when we stringify it to gets its representation, filter out these special dunder values such as __root__, unless they're at the top of the stack. This would allow us to see stats for __root__ still (is it useful?), whilst removing it from everything else.

jb098:
I knew this would be coming and didn't think it would be a concern. Having a transaction nesting everything kind of makes sense to me, in cases where you don't specify. However, __root__ really is not a very helpful. I wonder if rather than remove it entirely, we could turn it into something more useful. I'm thinking utlising __name__ from journey?

jarnfield:
Though suddenly, this does seem relevant to our conversation in slack. A non-string-building approach would give us more flexibility in scenarios such as this?

```
with context.tx("root"):
    with context.txn_tags(datapool="small"):
        with context.tx("userstore get"):
            .... 

>>> root :: userstore get datapool=small
```

cosmaprc (approved):

aecay:
#207 supersedes this (just in time for me to close it out for my last day 😅)

PR #161, opened 2020-10-28
Fix bug with header handling in response

jb098:
Mmeant we did not match RFC. Also add unit test.

### DIFF
--- a/acurl/acurl/__init__.py
+++ b/acurl/acurl/__init__.py
@@ -1,12 +1,14 @@
-import _acurl
-import threading
 import asyncio
-import ujson
 import shlex
+import threading
 import time
+from collections import defaultdict
 from urllib.parse import urlparse
-from pkg_resources import get_distribution, DistributionNotFound
 
+import ujson
+from pkg_resources import DistributionNotFound, get_distribution
+
+import _acurl
 
 try:
     __version__ = get_distribution(__name__).version
@@ -379,12 +381,9 @@ class Response:
     @property
     def headers(self):
         if not hasattr(self, '_headers'):
-            self._headers = dict()
+            self._headers = defaultdict(list)
             for k, v in self.headers_tuple:
-                if k in self._headers:
-                    self._headers[k].append(v)
-                else:
-                    self._headers[k] = [v]
+                self._headers[k].append(v)
             self._headers = {k: ', '.join(v) for k, v in self._headers.items()}
         return self._headers
 
--- a/acurl/acurl/__init__.py
+++ b/acurl/acurl/__init__.py
@@ -2,13 +2,13 @@ import asyncio
 import shlex
 import threading
 import time
-from collections import defaultdict
 from urllib.parse import urlparse
 
 import ujson
 from pkg_resources import DistributionNotFound, get_distribution
 
 import _acurl
+from acurl.utils import CaseInsensitiveDefaultDict
 
 try:
     __version__ = get_distribution(__name__).version
@@ -381,10 +381,11 @@ class Response:
     @property
     def headers(self):
         if not hasattr(self, '_headers'):
-            self._headers = defaultdict(list)
+            self._headers = CaseInsensitiveDefaultDict(list)
             for k, v in self.headers_tuple:
                 self._headers[k].append(v)
-            self._headers = {k: ', '.join(v) for k, v in self._headers.items()}
+            for k in self._headers:
+                self._headers[k] = ', '.join(self._headers[k])
         return self._headers
 
     # TODO: is this part of the request api?
--- a/acurl/acurl/__init__.py
+++ b/acurl/acurl/__init__.py
@@ -379,7 +379,13 @@ class Response:
     @property
     def headers(self):
         if not hasattr(self, '_headers'):
-            self._headers = dict(self.headers_tuple)
+            self._headers = dict()
+            for k, v in self.headers_tuple:
+                if k in self._headers:
+                    self._headers[k].append(v)
+                else:
+                    self._headers[k] = [v]
+            self._headers = {k: ', '.join(v) for k, v in self._headers.items()}
         return self._headers
 
     # TODO: is this part of the request api?
--- a/acurl/acurl/__init__.py
+++ b/acurl/acurl/__init__.py
@@ -8,7 +8,7 @@ import ujson
 from pkg_resources import DistributionNotFound, get_distribution
 
 import _acurl
-from acurl.utils import CaseInsensitiveDefaultDict
+from acurl.utils import CaseInsensitiveDefaultDict, CaseInsensitiveDict
 
 try:
     __version__ = get_distribution(__name__).version
@@ -381,11 +381,12 @@ class Response:
     @property
     def headers(self):
         if not hasattr(self, '_headers'):
-            self._headers = CaseInsensitiveDefaultDict(list)
+            headers = CaseInsensitiveDefaultDict(list)
             for k, v in self.headers_tuple:
-                self._headers[k].append(v)
-            for k in self._headers:
-                self._headers[k] = ', '.join(self._headers[k])
+                headers[k].append(v)
+            self._headers = CaseInsensitiveDict()
+            for k in headers:
+                self._headers[k] = ', '.join(headers[k])
         return self._headers
 
     # TODO: is this part of the request api?
--- a/acurl/acurl/utils.py
+++ b/acurl/acurl/utils.py
@@ -1,54 +1,52 @@
 from collections import defaultdict
 
 
-class CaseInsensitiveDefaultDict(defaultdict):
+class _CaseInsensitiveDict:
     # Base on https://stackoverflow.com/a/32888599 but tweaked for python3
-    # and using defaultdict as a base
-    @classmethod
-    def _k(cls, key):
+    @staticmethod
+    def _k(key):
         return key.lower() if isinstance(key, str) else key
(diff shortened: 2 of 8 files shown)


acurl/acurl/__init__.py
```diff
@@ -379,7 +379,13 @@ def json(self):
     @property
     def headers(self):
         if not hasattr(self, '_headers'):
-            self._headers = dict(self.headers_tuple)
+            self._headers = dict()
```
aecay, line 382: you could use `defaultdict(list)` here and decomplicate some of the `if k in d` logic.

aecay (approved):
one small suggestion, lgtm in general

acurl/acurl/__init__.py
jb098: Great suggestion, changing now.

jarnfield (approved):

acurl/acurl/__init__.py
```diff
@@ -379,7 +381,10 @@ def json(self):
     @property
     def headers(self):
         if not hasattr(self, '_headers'):
-            self._headers = dict(self.headers_tuple)
+            self._headers = defaultdict(list)
```
jarnfield, line 384: I'd probably be tempted to use an intermediate variable here rather than use self._headers and then rewrite it on line 387. I was a little confused at first glance.

acurl/acurl/__init__.py
```diff
@@ -379,7 +381,10 @@ def json(self):
...
+            for k, v in self.headers_tuple:
+                self._headers[k].append(v)
+            self._headers = {k: ', '.join(v) for k, v in self._headers.items()}
```
jarnfield, line 387:
```suggestion
            header = defaultdict(list)
            for k, v in self.headers_tuple:
                header[k].append(v)
            self._headers = {k: ', '.join(v) for k, v in header.items()}
```

acurl/acurl/utils.py
```diff
@@ -0,0 +1,54 @@
+from collections import defaultdict
+
+
+class CaseInsensitiveDefaultDict(defaultdict):
+    # Base on https://stackoverflow.com/a/32888599 but tweaked for python3
+    # and using defaultdict as a base
+    @classmethod
+    def _k(cls, key):
```
aecay, line 8:
```suggestion
    @staticmethod
    def _k(key):
```

acurl/acurl/utils.py
```diff
@@ -0,0 +1,54 @@
...
+        return key.lower() if isinstance(key, str) else key
+
+    def __init__(self, *args, **kwargs):
+        super(CaseInsensitiveDefaultDict, self).__init__(*args, **kwargs)
```
aecay, line 12:
```suggestion
        super().__init__(*args, **kwargs)
```

acurl/acurl/utils.py
```diff
@@ -0,0 +1,54 @@
...
+        self._convert_keys()
+
+    def __getitem__(self, key):
+        return super(CaseInsensitiveDefaultDict, self).__getitem__(self.__class__._k(key))
```
aecay, line 16:
```suggestion
        return super(CaseInsensitiveDefaultDict, self).__getitem__(self._k(key))
```

acurl/acurl/utils.py
```diff
@@ -0,0 +1,54 @@
...
+    def __delitem__(self, key):
+        return super(CaseInsensitiveDefaultDict, self).__delitem__(self.__class__._k(key))
+
+    def __contains__(self, key):
+        return super(CaseInsensitiveDefaultDict, self).__contains__(
+            self.__class__._k(key)
+        )
+
+    def has_key(self, key):
+        return super(CaseInsensitiveDefaultDict, self).has_key(self.__class__._k(key))
```
aecay, line 30: I'm a bit confused about this one because it's not an api method for dicts in python...

acurl/acurl/utils.py
```diff
@@ -0,0 +1,54 @@
...
+        )
+
+    def setdefault(self, key, *args, **kwargs):
+        return super(CaseInsensitiveDefaultDict, self).setdefault(
+            self.__class__._k(key), *args, **kwargs
+        )
+
+    def update(self, E={}, **F):
+        super(CaseInsensitiveDefaultDict, self).update(self.__class__(E))
+        super(CaseInsensitiveDefaultDict, self).update(self.__class__(**F))
```
aecay, line 49:
```suggestion
        super(CaseInsensitiveDefaultDict, self).update(self.__class__(E, **F))
```

I think we can get away with just a single call like this.

aecay (commented):
Some comments.  We should unit test the crap out of this too, but not necessarily for this PR since we're trying to get it in (tho core platform are holding their change til we give them the go-ahead...so maybe yes as part of this PR...)

acurl/acurl/utils.py
```diff
@@ -0,0 +1,54 @@
...
+            self.__class__._k(key)
+        )
+
+    def has_key(self, key):
+        return super(CaseInsensitiveDefaultDict, self).has_key(self.__class__._k(key))
+
+    def pop(self, key, *args, **kwargs):
+        return super(CaseInsensitiveDefaultDict, self).pop(
+            self.__class__._k(key), *args, **kwargs
+        )
```
aecay, line 35: We also need to implement popitem from the mutablemapping api: https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes

acurl/acurl/utils.py
jarnfield: Looks like it was removed some time between 2 and 3. I didn't even know it had gone.

aecay: I don't even know what python 2 is any more 😝

jb098:
> Some comments. We should unit test the crap out of this too, but not necessarily for this PR since we're trying to get it in (tho core platform are holding their change til we give them the go-ahead...so maybe yes as part of this PR...)

I agree on the unit test front. I have done some manual testing and there is a sort of higher level unit test that will test what I hope are any use cases that we're likely to see. I hope that gives us enough coverage for now.

I'll look at implementing your other changes.

acurl/acurl/utils.py
jarnfield: This seems like a good place to put my comment (thanks Aaron?)

Do we need to implement the full mutable mapping interface? I wonder what this implementation would look like as an abc.Mapping (none of the mutable stuff), that backed on to a dict, rather than extending it. Is there a use case for mutating response.headers?

acurl/acurl/__init__.py
```diff
@@ -379,7 +381,11 @@ def json(self):
     @property
     def headers(self):
         if not hasattr(self, '_headers'):
-            self._headers = dict(self.headers_tuple)
+            self._headers = CaseInsensitiveDefaultDict(list)
+            for k, v in self.headers_tuple:
+                self._headers[k].append(v)
+            for k in self._headers:
+                self._headers[k] = ', '.join(self._headers[k])
```
jarnfield, line 388: This is maybe slightly funky in that we're writing lists to self._headers and then rewriting them as strings, which might fly against the expectations set on line 384. Kind of relates to my first comment on the PR and could be worked around by applying the same change.

jb098: I did read your comment and though it was good in the first iteratios but I don't know why I just felt it was unpalatable to have multiple instances of this new type of dict floating about, purely because it needs that behaviour in both loops(maybe?). Although only the first one needs to be based off default dict I think? But that involves me writing more code to have two types of case insenstive ones.

However I will make two if people think that's better.

aecay: Ooh, yeah, I think having two is going to be better (with the first a local).  Only the second actually needs to be case-insensitive (and need not inherit from defaultdict either...)

jarnfield: That's fair. Having to have the case sensitivity logic in two places isn't something I'd considered.

acurl/acurl/utils.py
aecay: No - in fact, I was going to suggest that it would be better to have a frozen mapping here - but I didn't want to turn this into a Christmas tree of improvement ideas (more than it already is).  But in the abstract -- yes an immutable interface would definitely be preferable.

acurl/acurl/utils.py
jb098: Yeah I just read that, it will be expunged from my next commit.

acurl/acurl/utils.py
jb098: And on top of this, I understand E should default to None as it is optional.

acurl/acurl/utils.py
jb098: I don't think there is a case for mutating it, not like in the request. I think to make sure this is correct later for use cases of this unrelated to response.headers, we should to more unit tests.

acurl/acurl/__init__.py
jb098: No I think you're wrong about that @aecay, I think both need to be case insensitive as the header_tuple could contain lower and upper case 'keys'. I know the right thing is to have one that inherits from default and one from dict but I'm just lazy, I'll go do it.

aecay: 🤦 you're right.

aecay (approved):
One small typo (arf 🐶), otherwise lgtm 🎉

jarnfield (approved):

PR #169, opened 2020-11-30
Selenium built in page load metrics and context manager

Matt2298:
-Added metrics for mite selenium tests, including both page load metrics and network resource metrics for downstream resources.

- Fixed a bug regarding the Histogram whereby the bin_counts was being used to iterate on and append the metrics. if all of the metric value fell in the infinite bin, this would mean there was nothing to iterate through and therefore would not append the metrics.

- Some mad formatting.

### DIFF
--- a/mite/logoutput.py
+++ b/mite/logoutput.py
@@ -4,38 +4,38 @@ import time
 
 class MsgOutput:
     def __init__(self):
-        self._logger = logging.getLogger('MSG')
+        self._logger = logging.getLogger("MSG")
 
     def process_message(self, msg):
-        stacktrace = msg.pop('stacktrace', None)
+        stacktrace = msg.pop("stacktrace", None)
         if stacktrace and self._logger.isEnabledFor(logging.WARNING):
-            message = msg.pop('message', None)
-            ex_type = msg.pop('ex_type', None)
-            start = "[%s] %.6f" % (msg.pop('type', None), msg.pop('time', None))
-            end = ', '.join("%s=%r" % (k, v) for k, v in sorted(msg.items()))
+            message = msg.pop("message", None)
+            ex_type = msg.pop("ex_type", None)
+            start = "[%s] %.6f" % (msg.pop("type", None), msg.pop("time", None))
+            end = ", ".join("%s=%r" % (k, v) for k, v in sorted(msg.items()))
             self._logger.warning(
                 "%s %s\n%s: %s\n%s", start, end, ex_type, message, stacktrace
             )
         elif self._logger.isEnabledFor(logging.DEBUG):
-            start = "[%s] %.6f" % (msg.pop('type', None), msg.pop('time', None))
-            end = ', '.join("%s=%r" % (k, v) for k, v in sorted(msg.items()))
+            start = "[%s] %.6f" % (msg.pop("type", None), msg.pop("time", None))
+            end = ", ".join("%s=%r" % (k, v) for k, v in sorted(msg.items()))
             self._logger.debug("%s %s", start, end)
 
 
 class DebugMessageOutput:
     def __init__(self):
-        self._logger = logging.getLogger('Debug Logger')
+        self._logger = logging.getLogger("Debug Logger")
 
     def process_message(self, message):
         text = message.get("text")
-        if message.get('type') == "debug_console_message" and text:
+        if message.get("type") == "debug_console_message" and text:
             self._logger.info(text)
 
 
 class HttpStatsOutput:
     def __init__(self, period=2):
         self._period = period
-        self._logger = logging.getLogger('Http Stats')
+        self._logger = logging.getLogger("Http Stats")
         self._start_t = None
         self._req_total = 0
         self._req_recent = 0
@@ -63,9 +63,9 @@ class HttpStatsOutput:
         return self._error_total
 
     def process_message(self, message):
-        if 'type' not in message:
+        if "type" not in message:
             return
-        msg_type = message['type']
+        msg_type = message["type"]
         t = time.time()
         if self._start_t is None:
             self._start_t = t
@@ -73,11 +73,11 @@ class HttpStatsOutput:
             dt = t - self._start_t
             self._resp_time_recent.sort()
             self._logger.info(
-                'Total> #Reqs:%d #Errs:%d', self._req_total, self._error_total
+                "Total> #Reqs:%d #Errs:%d", self._req_total, self._error_total
             )
             self._logger.info(
-                'Last %d Secs> #Reqs:%d #Errs:%d Req/S:%.1f min:%s 25%%:%s 50%%:%s '
-                '75%%:%s 90%%:%s 99%%:%s 99.9%%:%s 99.99%%:%s max:%s',
+                "Last %d Secs> #Reqs:%d #Errs:%d Req/S:%.1f min:%s 25%%:%s 50%%:%s "
+                "75%%:%s 90%%:%s 99%%:%s 99.9%%:%s 99.99%%:%s max:%s",
                 self._period,
                 self._req_recent,
                 self._error_recent,
@@ -96,14 +96,18 @@ class HttpStatsOutput:
             del self._resp_time_recent[:]
             self._req_recent = 0
             self._error_recent = 0
-        if msg_type == 'http_metrics':
-            self._resp_time_recent.append(message['total_time'])
+        if msg_type == "http_metrics":
+            self._resp_time_recent.append(message["total_time"])
             self._req_total += 1
             self._req_recent += 1
-        elif msg_type == 'http_selenium_metrics':
-            self._resp_time_recent.append(message['total_time'])
+        elif msg_type == "http_selenium_page_load_metrics":
+            self._resp_time_recent.append(message["total_time"])
             self._req_total += 1
             self._req_recent += 1
-        elif msg_type in ('error', 'exception'):
+        elif msg_type == "http_selenium_network_resource_metrics":
+            self._resp_time_recent.append(message["total_time"])
+            self._req_total += 1
+            self._req_recent += 1
+        elif msg_type in ("error", "exception"):
             self._error_total += 1
             self._error_recent += 1
--- a/mite/logoutput.py
+++ b/mite/logoutput.py
@@ -96,15 +96,11 @@ class HttpStatsOutput:
             del self._resp_time_recent[:]
             self._req_recent = 0
             self._error_recent = 0
-        if msg_type == "http_metrics":
-            self._resp_time_recent.append(message["total_time"])
-            self._req_total += 1
-            self._req_recent += 1
-        elif msg_type == "http_selenium_page_load_metrics":
-            self._resp_time_recent.append(message["total_time"])
-            self._req_total += 1
-            self._req_recent += 1
-        elif msg_type == "http_selenium_network_resource_metrics":
+        if msg_type in {
+            "http_metrics",
+            "http_selenium_page_load_metrics",
+            "http_selenium_network_resource_metrics",
+        }:
             self._resp_time_recent.append(message["total_time"])
             self._req_total += 1
             self._req_recent += 1
--- a/mite_selenium/__init__.py
+++ b/mite_selenium/__init__.py
@@ -154,7 +154,7 @@ class _SeleniumWrapper:
     def get_js_metrics_context(self):
         return JsMetricsContext(self)
 
-    def wait(self, locator, timeout=5):
+    def wait_for_element(self, locator, timeout=5):
         try:
             return WebDriverWait(self._remote, timeout).until(
                 EC.presence_of_element_located(locator)
@@ -165,7 +165,10 @@ class _SeleniumWrapper:
             ) from te
 
     def find_element(self, locator):
-        return self.wait(locator, 2)
+        return self.wait_for_element(locator, 2)
+
+    def implicit_wait(self, timeout):
+        self._remote.implicitly_wait(timeout)
 
 
 class JsMetricsContext:
@@ -173,11 +176,11 @@ class JsMetricsContext:
         self._browser = browser
         self._results = None
 
-    async def __aenter__(self):
-        self._browser._clear_resource_timings()
+        async def __aenter__(self):
+            self._browser._clear_resource_timings()
 
-    async def __aexit__(self, exc_type, exc, tb):
-        self._results = self._browser._retrieve_javascript_metrics()
+        async def __aexit__(self, exc_type, exc, tb):
+            self._results = self._browser._retrieve_javascript_metrics()
 
 
 @asynccontextmanager
--- a/mite_selenium/__init__.py
+++ b/mite_selenium/__init__.py
@@ -1,6 +1,7 @@
 import logging
 from contextlib import asynccontextmanager
 
+from selenium.common.exceptions import TimeoutException
 from selenium.webdriver import Remote
 from selenium.webdriver.support import expected_conditions as EC
 from selenium.webdriver.support.ui import WebDriverWait
@@ -130,10 +131,7 @@ class _SeleniumWrapper:
         return converted_metrics
 
     def _convert_ms_to_seconds(self, value_ms):
-        try:
-            return value_ms / 1000
-        except Exception:
-            return value_ms
+        return value_ms / 1000
 
     def _retrieve_javascript_metrics(self):
         try:
@@ -145,6 +143,7 @@ class _SeleniumWrapper:
             return []
 
     def _clear_resource_timings(self):
+        breakpoint()
         self._remote.execute_script("performance.clearResourceTimings()")
 
     def get(self, url):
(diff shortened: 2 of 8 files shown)


mite_selenium/__init__.py
```diff
@@ -55,30 +55,43 @@ def stop(self):
...
         )
         return
 
     timings = context.browser.execute_script(
         'return performance.getEntriesByType("navigation")'
     )[0]
+    request_response_time = timings["responseEnd"] - timings["requestStart"]
+    first_contentful_paint = context.browser.execute_script(
+        'return performance.getEntriesByName("first-contentful-paint")'
+    )[0]["startTime"]
```
aecay, line 71: Are you sure this works?  I've tried it in the browser console, and it doesn't seem to return anything (neither for FF nor chrome).  https://web.dev/fcp/ seems to suggest a slightly more complicated arrangement is needed to get the info (need to register a performance event watcher for paint events, which otherwise would not be recorded, which makes sense in a way -- why would the browser capture high-density perf events on the off chance that someone might want them later).  See also [this lib function](https://github.com/GoogleChrome/web-vitals/blob/master/src/getFCP.ts).

(But if it wasn't working I guess we'd get KeyErrors or something...mostly I'm curious to see what metrics are returned here, in order to analyze them like I did for the navigation ones...)

mite_selenium/__init__.py
```diff
@@ -55,30 +55,43 @@ def stop(self):
...
-        tls_time=timings["secureConnectionStart"],
-        dns_time=timings["domainLookupEnd"],
-        connect_time=timings["connectEnd"],
-        total_time=timings["duration"],
+        dom_complete=_convert_ms_to_seconds(timings["domComplete"]),
+        dom_interactive=_convert_ms_to_seconds(timings["domInteractive"]),
+        transfer_size=_convert_ms_to_seconds(timings["transferSize"]),
+        transfer_start_timings=_convert_ms_to_seconds(timings["requestStart"]),
+        tls_time=_convert_ms_to_seconds(timings["secureConnectionStart"]),
+        dns_time=_convert_ms_to_seconds(timings["domainLookupEnd"]),
```
aecay, line 81:
```suggestion
        dns_time=_convert_ms_to_seconds(timings["domainLookupEnd"] - timings["domainLookupStart"]),
```

mite_selenium/__init__.py
```diff
@@ -55,30 +55,43 @@ def stop(self):
...
+        connect_time=_convert_ms_to_seconds(timings["connectEnd"]),
```
aecay, line 82:
```suggestion
        connect_time=_convert_ms_to_seconds(timings["connectEnd"] - timings["connectStart"]),
```

mite_selenium/__init__.py
```diff
@@ -55,30 +55,43 @@ def stop(self):
...
-        transfer_start_timings=["requestStart"],
-        tls_time=timings["secureConnectionStart"],
-        dns_time=timings["domainLookupEnd"],
-        connect_time=timings["connectEnd"],
-        total_time=timings["duration"],
+        dom_complete=_convert_ms_to_seconds(timings["domComplete"]),
+        dom_interactive=_convert_ms_to_seconds(timings["domInteractive"]),
+        transfer_size=_convert_ms_to_seconds(timings["transferSize"]),
+        transfer_start_timings=_convert_ms_to_seconds(timings["requestStart"]),
+        tls_time=_convert_ms_to_seconds(timings["secureConnectionStart"]),
```
aecay, line 80: I'm not totally sure what to do about these two (but they are not correct as written -- they should be somethingEnd - somethingStart in general).  There are three metrics, connectStart, connectEnd, and secureConnectionStart.  I think the way it breaks down is that connectStart until secureConnectionStart is the TCP connection overhead (`tcp_time` maybe), and secureConnectionStart to connectEnd is the TLS overhead (`tls_time`).  But it might be the case that the connection is not TLS (which we can get by inspecting timings["name"]), in which case we should just use connectStart - connectEnd and set the tls_time to zero.

There are also some exotic new protocols (http/3, sometimes also called SPDY or QUIC) that frontload some of the TLS negotiation overhead into the layer4 negotiation (which is not TCP any more, but a new protocol).  As long as we're talking directly to boring http servers inside the platform, then this will work fine.  But if we start talking to e.g. akamai, then (as they upgrade their systems) the browser might autonegotiate something exciting.  We can check nextHopProtocol and make sure it's equal to http/1.1 -- if not we probably want to log a warning that the more detailed assumptions about how the timing breaks down might not be valid.

mite_selenium/__init__.py
```diff
@@ -55,30 +55,43 @@ def stop(self):
...
         return
 
     timings = context.browser.execute_script(
         'return performance.getEntriesByType("navigation")'
     )[0]
+    request_response_time = timings["responseEnd"] - timings["requestStart"]
+    first_contentful_paint = context.browser.execute_script(
+        'return performance.getEntriesByName("first-contentful-paint")'
+    )[0]["startTime"]
+    time_to_interactive = timings["domComplete"] - first_contentful_paint
```
aecay, line 72:
```suggestion
    time_to_interactive = timings["domInteractive"] - first_contentful_paint
```
perhaps a small nit...I'm not totally sure what we are measuring here.  Intuitively it seems like "time that the user can see but not interact with the UI" which seems useful but idk (because I haven't looked) how it fits into accepted metrics of js perf.

aecay (commented):
The title of this PR will be a commit in this (open-source) repo, so we should make it be a publicly meaningful title and not an internal ticket number.

I've also made some comments about the selenium metrics we have -- I appreciate that these predate your PR, but we might as well fix them while we're here....

mite_selenium/__init__.py
```diff
@@ -51,32 +56,119 @@ def start(self):
...
+        if self._is_using_tls(timings["name"]):
+            return timings["secureConnectionStart"] - timings["connectStart"]
+        else:
+            return timings["connectEnd"] - timings["connectStart"]
+
+    def _send_page_load_metrics(self):
+        if self._browser_has_timing_capabilities():
+            timings, *_ = self._context.browser.execute_script(
+                "return performance.getEntriesByType('navigation')"
+            )
```
aecay, line 82: I'm pretty sure that this should just be a one-element list.  But we might want to check that and warn if there are more elements.  I'm not totally sure the circumstances under which the perf event buffer is reset, but in principle it could happen that we do `get(foo.com); get(bar.com)` and then the list of navigation perf events is `[foo, bar]` and we mistakenly report the foo stats for bar.

At least until we nail down how this all works, I'd prefer to be defensive here.

mite_selenium/__init__.py
```diff
@@ -51,32 +56,119 @@ def start(self):
...
+            if protocol != "http/1.1":
+                logger.warning(
+                    f"Timings may be inaccurate as protocol is not http/1.1: {protocol}"
+                )
+            metrics = {
+                "tcp_time": self._get_tcp_timing(timings),
+                "tls_time": self._get_tls_time_if_applicable(timings),
+                "dns_lookup_time": timings["domainLookupEnd"]
+                - timings["domainLookupStart"],
+                "time_to_first_byte": timings["responseStart"] - timings["connectionEnd"],
```
aecay, line 93: I think it's spelled `connectEnd`; at least that's what I've got in my browser console...

mite_selenium/__init__.py
```diff
@@ -51,32 +56,119 @@ def start(self):
...
+                "time_to_last_byte": timings["responseEnd"] - timings["connectionEnd"],
+                "page_weight": timings["transferSize"],
```
aecay, line 95: We have some dom timings we can put in here as well.  I think some subset of:
- domInteractive - requestStart (total time taken to load the page from an app perspective, i.e. excluding TCP and TLS negotiation)
- domInteractive (total time to interactive)
- domInteractive - responseEnd (additional time taken after the network finishes to prep the page -- incentivizes e.g. async loading of scripts so you don't block until the whole page is loaded to start loading additional resources)
- domContentLoadedEventEnd - domContentLoadedEventStart (time spent in JS attached to the onload event.  It's nice that we have timing for this specifically, and I think a good chunk of JS stuff does tend to hang off this event.  However it's partially an architectural question: you could probably build an app that puts its processing elsewhere, to the usefulness will depend on the specifics of the design)
- There are also metrics for the unload event fired in the previous page (before navigating).  But that's even more architecture-dependent and less likely to be useful (and I don't think we need to bother with it, probably...)

mite_selenium/__init__.py
```diff
@@ -51,32 +56,119 @@ def start(self):
...
+            }
+            self._context.send(
+                "http_selenium_metrics",
+                *self._extract_and_convert_metrics_to_seconds(metrics),
+            )
+
+    def _extract_and_convert_metrics_to_seconds(self, metrics):
+        converted_metrics = defaultdict()
```
aecay, line 103:
```suggestion
        converted_metrics = dict()
```

Unless I'm missing something, defaultdict isn't needed here...

mite_selenium/__init__.py
```diff
@@ -51,32 +56,119 @@ def start(self):
...
+        for k, v in metrics:
+            converted_metrics[k] = self._convert_ms_to_seconds(v)
+        return converted_metrics
+
+    # TODO: Implement
+    def _send_javascript_metrics(self):
+        pass
+
+    def _convert_ms_to_seconds(value_ms):
+        return (value_ms / 1000) % 60
```
aecay, line 113: Why modulo 60?

mite_selenium/__init__.py
```diff
@@ -51,32 +56,119 @@ def start(self):
...
+    async def publish_javascript_metrics(self, transaction_name):
+        try:
+            print("Doing something for transaction")
+            yield
+        finally:
+            self._send_javascript_metrics()
+
+    def get(self, url):
+        self._context.browser.get(url)
+        self._send_page_load_metrics()
```
aecay, line 125:
```suggestion
    async def get(self, url):
        await self._context.browser.get(url)
        self._send_page_load_metrics()
```

This is how it *should* work...I hope it's true (if not we have async api design problems)

Matt2298: I've changed this to be async 👍

mite_selenium/__init__.py
Matt2298: I had something in the back of my mind that had me thinking I needed to make sure to modulo 60. There's no need for this.

mite_selenium/__init__.py
Matt2298: Misunderstood the use case for defaultdict, changed this now

mite_selenium/__init__.py
Matt2298: I've added these suggested metrics to the list we collect, however, the namings might need reviewing.

mite_selenium/__init__.py
Matt2298: Indeed it is, good spot 👍

mite_selenium/__init__.py
Matt2298: I'm not quite sure how you mean? In this instance from my understanding, we're taking the head element as timings and ignoring the rest. Is the suggestion to change this to use the following:

`timings = self._context.browser.execute_script(...)[0]`

mite_selenium/__init__.py
```diff
@@ -51,32 +56,119 @@ def start(self):
...
+        return self._context.browser.capabilities["browserName"] != "chrome"
+
+    def _is_using_tls(self, name):
+        return name.startsWith("https")
+
+    def _get_tls_timing(self, timings):
+        if self._is_using_tls(timings["name"]):
+            return timings["connectEnd"] - timings["secureConnectionStart"]
+        else:
+            logger.warning("Secure TLS connection not used, defaulting tls_time to 0")
```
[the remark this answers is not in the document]
Matt2298: Good point, changed this to info

mite_selenium/__init__.py
```diff
@@ -51,32 +56,119 @@ def start(self):
...
-        transfer_size=timings["transferSize"],
-        transfer_start_timings=["requestStart"],
-        tls_time=timings["secureConnectionStart"],
-        dns_time=timings["domainLookupEnd"],
-        connect_time=timings["connectEnd"],
-        total_time=timings["duration"],
-        effective_url=timings["name"],
-    )
+    def _browser_has_timing_capabilities(self):
+        return self._context.browser.capabilities["browserName"] != "chrome"
```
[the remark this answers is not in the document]
Matt2298: From what I have seen, I can't access paint timings in firefox, whereas using chrome I can see the paint perf entries. I think for now, we should keep this the same and then raise a ticket to add proper handling of the capabilities based on the browser. What do you think? I think there's quite a bit of work involved around making it generic enough to support multiple browsers.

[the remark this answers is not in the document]
aecay: I agree, let's make it a tech debt ticket on jira, and maybe also an issue in this open source repo.

mite_selenium/__init__.py
aecay: `performance.getEntriesByType('navigation')` returns a list of page navigation timings.  We are assuming that the zeroth element of that list is relevant to the current page.  That's probably correct if we load the page by (a simulated version of) typing the address into the address bar and hitting enter.  There are various JS apis to navigate though that a page could call.  I'm not sure whether all of those would also clear the list, or whether the list would survive across those.  So we might get into a situation where the list returned by `performance.getEntriesByType('navigation')` has multiple entries.  (There's then a subquestion about whether new entries are added to the beginning or the end of the list, but I assume it's add-to-end).

So, we have a packet of assumptions about this list being exactly one element long; if those assumptions are violated we'll get an error -- either a runtime error in python if there are zero elements, or a logic error (recording the wrong timings) if there are more.  I'm saying we should have an explicit test of the returned list length, rather than just assuming it will always be one, and we should log a specific error if that's not fulfilled.

Matt2298: I've wrapped this to make sure that we're explicitly only acting on one list item

mite/logoutput.py
```diff
@@ -96,14 +96,18 @@ def process_message(self, message):
...
             self._req_total += 1
             self._req_recent += 1
-        elif msg_type == 'http_selenium_metrics':
-            self._resp_time_recent.append(message['total_time'])
+        elif msg_type == "http_selenium_page_load_metrics":
+            self._resp_time_recent.append(message["total_time"])
             self._req_total += 1
             self._req_recent += 1
-        elif msg_type in ('error', 'exception'):
+        elif msg_type == "http_selenium_network_resource_metrics":
```
aecay, line 107: `if msg_type in {"http_metrics", ...}`

mite_selenium/__init__.py
```diff
@@ -48,45 +52,145 @@ def start(self):
...
+                converted_metrics[k] = self._convert_ms_to_seconds(v)
+            else:
+                converted_metrics[k] = v
+        return converted_metrics
+
+    def _convert_ms_to_seconds(self, value_ms):
+        try:
+            return value_ms / 1000
+        except Exception:
+            return value_ms
```
aecay, line 136: Hmm.  What exceptions have you seen here?  This seems like an overly permissive code path.

mite_selenium/__init__.py
```diff
@@ -48,45 +52,145 @@ def start(self):
...
+
+    def get_js_metrics_context(self):
+        return JsMetricsContext(self)
+
+    def wait_for_element(self, locator, timeout=5):
+        try:
+            return WebDriverWait(self._remote, timeout).until(
+                EC.presence_of_element_located(locator)
+            )
+        except Exception as te:
```
aecay, line 162: I think you need to catch something more specific than just `Exception` here, otherwise, if there is some other sort of error (process interrupted?) then the logs will misleadingly say that there was a timeout

mite_selenium/__init__.py
```diff
@@ -48,45 +52,145 @@ def start(self):
...
+            raise MiteError(
+                f"Timed out trying to find element '{locator}' in the dom"
+            ) from te
+
+    def find_element(self, locator):
+        return self.wait_for_element(locator, 2)
```
aecay, line 168: What is the point of this method; can't people just call `wait_for_element` directly?

mite_selenium/stats.py
```diff
@@ -1,15 +1,59 @@
...
-    ),
     Counter(
-        'mite_http_selenium_response_total',
-        matcher_by_type('http_selenium_metrics'),
-        extractor('test journey transaction'.split()),
+        "mite_http_selenium_response_total",
+        matcher_by_type("http_selenium_page_load_metrics"),
+        extractor("test journey transaction".split()),
+    ),
+    *build_metrics(_PAGE_LOAD_METRICS, "http_selenium_page_load", ["transaction"]),
```
aecay, line 53:
```suggestion
    *build_metrics(_PAGE_LOAD_METRICS, "selenium_page_load", ["transaction"]),
```

I think `mite_selenium` probably makes better sense as a metric prefix here (generally, not just on this specific line).  The inclusion of http here is probably a carryover from the metrics from the mite_http module / acurl.  But these metric names are already very very long so whatever savings we can make are worth it....

test/test_webdriver.py
```diff
@@ -86,7 +86,7 @@ async def test(context):
         pass
 
     # patch with async decorator misbehaving
-    with patch('mite_selenium.Remote', autospec=True) as MockRemote:
+    with patch("mite_selenium.Remote", autospec=True) as MockRemote:
```
aecay, line 89: Unit tests for the new api surfaces? 🙏

mite/logoutput.py
Matt2298: Not sure why I didn't clock that, updated this

mite_selenium/__init__.py
Matt2298: I think originally my thinking was around converting values that weren't meant to be converted to seconds, as I was lazily operating on all values. I had changed it since to not call the convert if nnot required so this can be simplified

mite_selenium/__init__.py
Matt2298: You're right, this needs to be a timeout exception hence the `te`, not sure why I left it as Exception

mite_selenium/__init__.py
Matt2298: yes we can, I think I wanted to reuse the wait side of it to make sure that the element existed, but wait does exactly that

mite_selenium/stats.py
Matt2298: I've reduced this down to `selenium_page_load` as suggested

test/test_webdriver.py
Matt2298: Added unit tests to cover the interface changes, thank you for the help on that

aecay (approved):

DavAnnunz (approved):

PR #170, opened 2020-12-01
Restore postmortem debugging

aecay:
It turns out that there are differences between set_trace and
post_mortem debugger calls.  Unfortunately PEP553 doesnʼt support
customizing the latter, so we have to DIY.


jb098 (approved):

DavAnnunz (approved):

paulpepper-sky (commented):
With side-channel explanation from Aaron, makes sense to default to using  `pdb.post_mortem()` as a way to enter the debugger at the stack frame from the exception, rather than at the point where the `breakpoint()` is called.

PR #175, opened 2021-01-06
Provide access to the Selenium webdriver instance on _SeleniumWrapper

paulpepper-sky:
These changes give access to the Selenium webdriver via context. Normally classes wrapped by _SeleniumWrapper proxy calls to webdriver methods through that class's methods - for instance to collect statistics. However, there are cases where proxied access isn't required or desirable.

### DIFF
--- a/mite_selenium/__init__.py
+++ b/mite_selenium/__init__.py
@@ -50,6 +50,7 @@ class _SeleniumWrapper:
             file_detector=self._file_detector,
             options=self._options,
         )
+        self._context.webdriver = self._remote
 
     def _stop(self):
         self._remote.close()
--- a/mite_selenium/__init__.py
+++ b/mite_selenium/__init__.py
@@ -50,7 +50,7 @@ class _SeleniumWrapper:
             file_detector=self._file_detector,
             options=self._options,
         )
-        self._context.webdriver = self._remote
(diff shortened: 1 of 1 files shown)


mite_selenium/__init__.py
```diff
@@ -50,6 +50,7 @@ def _start(self):
             file_detector=self._file_detector,
             options=self._options,
         )
+        self._context.webdriver = self._remote
```
aecay, line 53: I think that we ought to signal somehow that this should not be the first port of call for users, either by calling it `_webdriver` or `internal_webdriver` or something like that.  (I actually don't see a huge issue with the status quo where it can be accessed as `context.browser._remote`, although I can also see why that's not optimal...)

paulpepper-sky: I had initially gone the route of accessing the driver via `context.browser._remote`, but as you say, not optimal - Demeter and all that. Neither does it seem like an internal only attribute, but as you also point out, it shouldn't be the primary choice for accessing the Webdriver instance. How about naming the instance `raw_webdriver` or something similar to convey that secondary use intent?

aecay: 👍 for raw_webdriver

jarnfield (approved):

aecay (approved):

PR #177, opened 2021-01-20
Test session cookies sent on subsequent requests

Matt2298:
- Added tests to cover scenario whereby cookies should be sent to domain on subsequent requests after being dropped on the initial request.

### DIFF
--- a/acurl/tests/test_httpbin.py
+++ b/acurl/tests/test_httpbin.py
@@ -40,6 +40,19 @@ async def test_session_cookies(httpbin):
 
 
 @pytest.mark.asyncio
+async def test_session_cookies_sent_on_subsequent_request(httpbin):
+    s = session()
+    await s.get(httpbin.url + "/cookies/set?name=value")
+    cookie_list = s.get_cookie_list()
+    assert len(cookie_list) == 1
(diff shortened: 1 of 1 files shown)


jarnfield (approved):

acurl/tests/test_httpbin.py
```diff
@@ -39,6 +39,19 @@ async def test_session_cookies(httpbin):
     assert len(cookie_list) == 0
 
 
+@pytest.mark.asyncio
+async def test_session_cookies_sent_on_subsequent_request(httpbin):
+    s = session()
+    await s.get(httpbin.url + "/cookies/set?name=value")
+    cookie_list = s.get_cookie_list()
+    assert len(cookie_list) == 1
+    assert cookie_list[0].name == "name"
```
aecay, line 48: These assertion duplicate what's in `test_session_cookies` so I'd tend to think they don't belong here.  Otherwise this lgtm.

aecay (approved):

acurl/tests/test_httpbin.py
Matt2298: Good point, removed these assertions

PR #178, opened 2021-01-21
JS context timing 

haffi96:
Add timer to JSMetricsContext

### DIFF
--- a/mite_selenium/stats.py
+++ b/mite_selenium/stats.py
@@ -23,7 +23,10 @@ _NETWORK_RESOURCE_METRICS = [
     ("time_to_last_byte", "seconds"),
     ("tls_time", "seconds"),
     ("total_time", "seconds"),
-    ("js_execution_time", "seconds")
+]
+
+_CUSTOM_METRICS = [
+    ("js_execution_time", "seconds"),
 ]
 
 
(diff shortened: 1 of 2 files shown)


mite_selenium/stats.py
```diff
@@ -23,6 +23,7 @@
     ("time_to_last_byte", "seconds"),
     ("tls_time", "seconds"),
     ("total_time", "seconds"),
+    ("js_execution_time", "seconds")
```
aecay, line 26: I commented on this in the other PR as well, but I'll bring it here too: `js_execution_time` isn't part of a network resource request; it's it's own top-level thing and so it needs a separate metric not one nested in here.

mite_selenium/stats.py
```diff
@@ -57,4 +58,5 @@ def build_metrics(metrics, matcher, labels):
         "selenium_network_resource",
         ["transaction", "resource_path"],
     ),
+    *build_metrics(_CUSTOM_METRICS, "selenium_custom", ["transaction"]),
```
aecay, line 61:
```suggestion
    *build_metrics(_CUSTOM_METRICS, "selenium", ["transaction"]),
```

aecay (approved):

DavAnnunz (approved):

PR #179, opened 2021-01-21
Fix to add missing perform_request() function prototype.

paulpepper-sky:
`acurl.h` was missing the `perform_request()` function prototype, causing a gcc build error in some (but not all) versions of gcc when running `pip install acurl`. This PR adds the prototype to `acurl.h` to fix that issue.


aecay (approved):

Matt2298 (approved):

paulpepper-sky:
acurl.h was missing the perform_request() function prototype, causing a gcc build error in some (but not all) versions of gcc when running pip install acurl. This PR adds the prototype to acurl.h to fix that issue.

PR #185, opened 2021-02-15
Fix HTTP 100 handling

jarnfield:


acurl/acurl/__init__.py
```diff
@@ -400,13 +401,28 @@ def headers(self):
                 self._headers[k] = ", ".join(headers[k])
         return self._headers
 
+    def _get_header_lines(self):
+        header_lines = iter(self.header.splitlines())
```
aecay, line 405: This should split on `'\r\n'` explicitly.  why?  we are parsing a protocol here, and we should be strict in what we match to avoid vulnerability to things like https://portswigger.net/web-security/request-smuggling -- that is a different example (revolving around content-encoding: chunked), but the point is that if we deviate in our interpretation of the protocol, we open up inconsistencies that are potentially exploitable.  splitlines allows splitting on some exotic ascii and nonascii characters.  (i think the security risk profile of acurl is vanishingly small, but that's no reason not to dtrt especially when it's relatively easy...)

we also don't need `x = iter(foo); for y in x: ...`; we can just skip straight to `for y in foo: ...`

acurl/acurl/__init__.py
```diff
@@ -400,13 +401,28 @@ def headers(self):
...
+        for line in header_lines:
```
aecay, line 406: This lets us have "100 continue" lines at any point in the response, right?  But they can only happen at the beginning, right?  It's also less efficient because it's doing a regex match against each line.

I'd rather we implemented this in a much simpler way:

```
headers = [l for l in self.headers.split("\r\n")]
headers = headers[:-2] # drop the final blank lines
if headers[0].startswith("HTTP/1.1 100"):  # http 1.0 doesn't define 100 continue; and we can be pretty sure there's no http/1.2 in the offing....
    headers = headers[2:]  # it's not clear to me if a 100 Continue response can have headers itself or what we'd do if it did
return headers
```

acurl/acurl/__init__.py
jarnfield: The iter() call was to enable line 411. But I think other points you raised negate the point anyway (and if there's a better way to do this that I've missed in my overconfidence, please do clarify further). No complains with the crlf 
change!

acurl/acurl/__init__.py
jarnfield: So, I went with the iterative approach because I wasn't sure if you could encounter multiple HTTP 100s before getting the final response code. Even so, we could certainly take your approach and adapt it to cope with many, with the addition of a single line. But I'm not sure if the spec's particularly clear about the possibility. Will use your version as writted for now and we can worry about it when it (probably never) happens:

https://stackoverflow.com/questions/22818059/several-100-continue-received-from-the-server

acurl/acurl/__init__.py
aecay: ah right, yeah that makes sense.  thanks for the 411 😝 (do people even say that in england?)

acurl/acurl/__init__.py
jarnfield: In fact it's not even a new line is it...

aecay: No newline, just a carriage return 😝

acurl/acurl/__init__.py
```diff
@@ -1,4 +1,5 @@
 import asyncio
+import re
```
aecay, line 2: unneeded import, now

aecay (approved):

cosmaprc (approved):

PR #186, opened 2021-02-17
Exploring unbalanced controller work allocation in cybertron 10M tests

aecay:
WIP branch, no functional changes for now but some code simplifications and we've raised the level of a crucial log message to INFO to help debugging


aecay:
not needed

PR #187, opened 2021-02-25
Fixes for reaching higher 7.5 million TPS during testing

cosmaprc:
Fix for Recyclable iterable datapool getting exhausted on higher 7.5 million TPS

Remove selenium from mite stats in an effort to lower the nr of stats replicas needed to reach 7.5 million TPS

### DIFF
--- a/mite/datapools.py
+++ b/mite/datapools.py
@@ -1,5 +1,6 @@
 import logging
 from collections import deque, namedtuple
+from itertools import cycle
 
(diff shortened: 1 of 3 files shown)


Matt2298 (approved):

aecay (changes requested):
we should not commit these changes to mite.  We need to make the appropriate modifications in id-mite-nft

cosmaprc:
Closing this PR as it's no longer needed

PR #188, opened 2021-02-25
make it possible to configure on the command line which processors will be used by mite stats

aecay:


cosmaprc (approved):

docs/stats.rst
```diff
@@ -111,3 +111,37 @@ package:
...
+
+Configuring custom stats
+------------------------
+
+By default, all stats processors that are registered as entrypoints will
+be used by the mite stats component.  However, as each processor has to
+execute code for each message (even if only to determine that it is not
+interested in that message), the performance impact of having many stats
+processors registered can be large.  The ``--stats-include-processors``
+and ``--stats-exclude-processors`` can be used to fine-tune this behavior.
```
jarnfield, line 124: Looks great. Just in case... feels like you maybe meant to put a word before "can", as you opened the sentence with "The". "switches" or something, maybe? If it makes enough sense to you, as it does to me, then you can safely ignore this comment.

jarnfield (approved):

docs/stats.rst
aecay: X = (the performance impact of (having many stats processors registered))
X can be large

That's what I meant to say, and I think it is "correct" as written (unless I'm missing something)...but if it tripped you up then whether it's "correct" or not it's unclear and thus not a good fit for documentation, so I will look to rephrase it.  Thanks.

aecay: > However, because each processor has to execute code for each message (even if only to determine that it is not interested in that message), there can be a large cumulative performance impact if you have many stats processors registered.

That's how it reads now which is hopefully an improvement

Matt2298 (approved):

PR #189, opened 2021-03-25
Add selenium switch_to_iframe

cosmaprc:


mite_selenium/__init__.py
```diff
@@ -162,6 +162,11 @@ def wait_for_element(self, locator, timeout=5):
                 f"Timed out trying to find element '{locator}' in the dom"
             ) from te
 
+    def switch_to_iframe(self, index):
+        self._remote.switch_to.frame(
+            self._remote.find_elements_by_tag_name("iframe")[index]
```
aecay, line 167: Offering only the choice to find the element by index is not the greatest -- we might have a page with many iframes, a changing number, etc.  Instead, this function should look like:
```
def switch_to_iframe(self, by, locator):
```

`by` should be one of the [selenium `By` values](https://selenium-python.readthedocs.io/locating-elements.html) and `locator` should be the corresponding tag name, css selector etc. to pass down to `driver.find_element`

paulpepper-sky: This XPath, or something like it, might do the job and it's self-documenting into the bargain. 
`//iframe[contains(@title, "iFrame containing Sky Sign-In application")]
`

cosmaprc: used this xpath, thanks

mite_selenium/__init__.py
```diff
@@ -162,10 +162,8 @@ def wait_for_element(self, locator, timeout=5):
                 f"Timed out trying to find element '{locator}' in the dom"
             ) from te
 
-    def switch_to_iframe(self, index):
-        self._remote.switch_to.frame(
-            self._remote.find_elements_by_tag_name("iframe")[index]
-        )
+    def switch_to_iframe(self, locator):
+        self._remote.switch_to.frame(self.wait_for_element(locator))
```
aecay, line 166: this has different semantics...is there a reason not to use find_element?

cosmaprc: used wait_for_element because it was already there and it has added timeout functionality and underneath it does `find_element(*by)` so I don't have to expand it.

cosmaprc: ok, good point, I didn't think about that. Have made the change to use find_element

aecay (approved):

paulpepper-sky (approved):

PR #191, opened 2021-04-07
Make mypy apply

aecay:
This isn't a joined-up attempt to add type annotations to the library, but rather the minimum (more or less) that we need to make mypy run over this code (and our own closed-source tests).  In the process we realized these simplifications:
- refactored `with_fixed_separation` and friends to use `contextlib`
- migrated the `Scenario` class to a nicer NamedTuple syntax
- improved the code in volume_model (broke one mega-function into slightly smaller pieces -- I'm not totally sure whether this fixed any bugs or not)

All the other changes are cosmetic

### DIFF
--- a/acurl/acurl/__init__.py
+++ b/acurl/acurl/__init__.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import asyncio
 import shlex
 import threading
--- a/acurl/acurl/__init__.py
+++ b/acurl/acurl/__init__.py
@@ -7,7 +7,7 @@ from urllib.parse import urlparse
 import ujson
 from pkg_resources import DistributionNotFound, get_distribution
 
-import _acurl
+import _acurl  # type: ignore
 from acurl.utils import CaseInsensitiveDefaultDict, CaseInsensitiveDict
 
 try:
@@ -375,7 +375,7 @@ class Response:
 
     # TODO: why do we allow setter?
     @encoding.setter
-    def encoding_setter(self, encoding):
+    def encoding(self, encoding):
         self._encoding = encoding
 
     @property
@@ -392,10 +392,12 @@ class Response:
     @property
     def headers(self):
         if not hasattr(self, "_headers"):
-            headers = CaseInsensitiveDefaultDict(list)
+            headers: CaseInsensitiveDefaultDict[
+                str, list[str]
+            ] = CaseInsensitiveDefaultDict(list)
             for k, v in self.headers_tuple:
                 headers[k].append(v)
-            self._headers = CaseInsensitiveDict()
+            self._headers: CaseInsensitiveDict[str, str] = CaseInsensitiveDict()
             for k in headers:
                 self._headers[k] = ", ".join(headers[k])
         return self._headers
--- a/acurl/acurl/utils.py
+++ b/acurl/acurl/utils.py
@@ -1,13 +1,10 @@
 from __future__ import annotations
 
 from collections import defaultdict
-from typing import MutableMapping, TypeVar
-
-_K = TypeVar("_K")
-_V = TypeVar("_V")
+from typing import MutableMapping
 
 
-class _CaseInsensitiveDict(MutableMapping[_K, _V]):
+class _CaseInsensitiveDict(MutableMapping):
     # Base on https://stackoverflow.com/a/32888599 but tweaked for python3
     @staticmethod
     def _k(key):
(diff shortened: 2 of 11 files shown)


jb098 (commented):
So I've remainded apprehensive about adding MyPy to mite, purely because I worry it increases the bar to entry for people in the open source community.

I am however open to being convinced, and I see that this doesn't introduce sweeping changes. It would be good if you could outline some of the positives you've seen by introducing it.

aecay:
The positives, unfortunately, are not mostly seen in this repo, at this point.  If we wanted that, we'd need to add more typings throughout the mite code.  Those would then serve as a safety net to ensure future changes didn't make the kinds of errors that the type checker prevents (forgetting to check for `None`, returning a string when you should have converted it to an integer, ...).  The type checker, even without type annotations, does protect against things like misspelled method names, imports, things like that.  I see those as pretty low-risk in the mite repo though -- because the test coverage here is pretty high (which is another way of catching those errors) and because our team don't tend to make changes in this repo unless we are going to turn right around and use those changes in our closed-source tests (which is another kind of verification).

Introducing this fixed one "bug" in the mite code -- the setter for the `encoding` property on one of the acurl classes had the wrong name.  I'm not sure if it would misbehave at runtime, but the [docs](https://docs.python.org/3/library/functions.html#property) say

> Be sure to give the additional functions the same name as the original property

So it must be important for some reason.  (I also actually suspect the real bug is making that property settable at all, but that's a different issue...)  Anyway, for better or for worse that's the (low) stakes that we are playing for in this repo.

The big(ger) benefit comes in our closed-source test suite, where the "high test coverage" and "all API surfaces exercised quite frequently" properties that hold in this repo are violated.  I found 3 functions that got deleted by git rebase mistakes, a handful of mistyped imports or property names, and a few other things.  These were all in the places in the code you might expect -- the poorly exercised ones.  (Idris tests, chaos tests and the annotations that go with them, bits of code to manually run certain journeys, where the underlying journeys then get changed, ...).  So the benefit there, as I see it, is not in ensuring the day-to-day correctness of our code (we already do that ourselves by noticing/fixing runtime errors), but in protecting the "long tail" from bitrot.

The annotations added here, to reiterate, are not what you'd call "proper" types for this repo; it's just (mostly) the bare minimum that we need to get mypy running cleanly over this code, in conjunction with the id-mite-nft closed-source repo.  I did take the attitude that if I was going to be making changes here, I would do more or less the "proper" thing, rather than the minimal thing (like slapping `Any` annotations on anything that needed it, or marking it all `# type: ignore`).  I could do that (and in fact could just mark mite/acurl as ignored from within the id-mite-nft repo, necessitating no changes here -- though that would slightly weaken our coverage in id-mite-nft).  I tried to tread a middle way...

Dunno if that answers your questions...

PS I (still) think that mypy annotations are a bad fit for mite overall because of some of the dynamic things we do that the type system supports poorly or not at all.  So I'm not planning to add a full complement of annotations, and I definitely wouldn't be bothering unless I had id-mite-nft in front of me.  I have come around to the idea that the check that you get "for (mostly) free" without (many) annotations are worthwhile in the id-mite-nft codebase.

PPS speaking of open-source contributors, I realize I haven't plumbed mypy checks up to the CI in this repo, only in id-mite-nft.  I'll do that...

jb098:
Initially I thought putting `Any` everywhere might be preferred but I think if you're going to do that, you might as well add some types. I'm still on the fence a little, I'd like to hear what others think. Maybe someone experienced like @johnarnfield or someone who has worked with mite for a long time like @DavAnnunz.

jb098:
It might be extra work but could we add 'typed wrappers' for want of a better term in our id-mite-nft repo to leave the orginal mite interface the same?

aecay:
Yeah -- we could create "interface" files for mite.  There's actually one of those in this PR already (acurl/acurl/utils.pyi) -- mostly because typing that module inline turned it into a pig's breakfast.  I would say that if we are doing that we should probably put the stubs in this repo (to avoid divergence, and potentially let other users of mite benefit from/improve on them) -- but that could be done without any changes in the `.py` files.

jb098:
> Yeah -- we could create "interface" files for mite. There's actually one of those in this PR already (acurl/acurl/utils.pyi) -- mostly because typing that module inline turned it into a pig's breakfast. I would say that if we are doing that we should probably put the stubs in this repo (to avoid divergence, and potentially let other users of mite benefit from/improve on them) -- but that could be done without any changes in the `.py` files.

Actually I quite like that as an idea, typed interfaces for the general public. I like to give the community options without forcing anything upon them However, it would mean more work to this PR. What do you think?

aecay:
I think I'd like to land the mypy changes for id-mite-nft (so we can start benefiting from them...)  So in the first instance I will mark mite and acurl as ignored for the typechecker in that repo, just to get that green and landable.  Then I'll follow on with type interface files for this repo -- that will probably be lower priority in the nature of things, but since mite is open source I also don't have to stop working on it after the 1st of June 😉

aecay:
I'll close this one for now

PR #192, opened 2021-04-09
Adds wait_for_elements method to mite selenium

Matt2298:
- Adding a new method for waiting for multiple elements on a page to appear.

### DIFF
--- a/mite_selenium/__init__.py
+++ b/mite_selenium/__init__.py
@@ -154,17 +154,20 @@ class _SeleniumWrapper:
         return JsMetricsContext(self)
 
     def wait_for_elements(self, locator, timeout=5):
+        return self._wait_for(EC.presence_of_all_elements_located, locator, timeout)
+
+    def wait_for_element(self, locator, timeout=5):
+        return self._wait_for(EC.presence_of_element_located, locator, timeout)
+
+    def _wait_for(self, condition_func, locator, timeout=5):
         try:
             return WebDriverWait(self._remote, timeout).until(
-                EC.presence_of_all_elements_located(locator)
+                condition_func(locator)
             )
         except TimeoutException as te:
             raise MiteError(
-                f"Timed out trying to find element(s) '{locator}' in the dom"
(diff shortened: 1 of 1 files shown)


jb098:
I'm wondering whether we need to add this to mite or whether it's more of a helper method we should have in our tests. It's definitely a re-usable piece of selenium code but I feel like we should keep the interface to between mite/selenium quite light where possible.

Matt2298:
> I'm wondering whether we need to add this to mite or whether it's more of a helper method we should have in our tests. It's definitely a re-usable piece of selenium code but I feel like we should keep the interface to between mite/selenium quite light where possible.

The web driver instance itself "isn't" (shouldn't be) accessible on the context so we expose methods in mite which make use of the web driver. This PR is just refactoring a method to be a bit more flexible than before. I don't foresee this changing much more than this

haffi96 (approved):

aecay (commented):
I have Jordan's question but in reverse 😝 I wonder, given that we can wait for elements, whether we need also the ability to wait for a single element (since that's equivalent to waiting for a 1-entry list).

I suspect the balance probably tips towards having both methods (the singular and the plural), but it's at least worth checking whether others agree. I do think that keeping the mite API surface as small as possible is a good thing (outweighed in this case by ergonomics, but still an active consideration)

WDYT?

mite_selenium/__init__.py
```diff
@@ -153,19 +153,20 @@ def get(self, url):
...
-            )
+            return WebDriverWait(self._remote, timeout).until(condition_func(locator))
         except TimeoutException as te:
             raise MiteError(
                 f"Timed out trying to find element '{locator}' in the dom"
             ) from te
 
-    def switch_to_iframe(self, locator):
-        self._remote.switch_to.frame(self._remote.find_element(*locator))
-
```
aecay, line 168: intentional removal?

aecay (approved):

mite_selenium/__init__.py
Matt2298: Nope 🤦

PR #193, opened 2021-04-16
IPB-7683 - extract timings from har file§

Sarah-J-S:
Changes added to extract timings from har file.

NB the sleep time for the first entry will always be zero as there is no preceding time. I have assumed that non-zero user supplied sleep times take precedence over captured times.

### DIFF
--- a/test/test_har_to_mite.py
+++ b/test/test_har_to_mite.py
@@ -1,68 +1,68 @@
-import pytest
 import os
-
 from copy import deepcopy
+
+import pytest
 
 from mite.har_to_mite import (
     _extract_and_sort_requests,
     _parse_urls,
     _render_journey_transaction,
+    har_convert_to_mite,
     set_expected_status_code,
-    set_request_headers_dict,
     set_request_body,
-    har_convert_to_mite
+    set_request_headers_dict,
 )
 
 # page with status_code=0
 SIMPLE_PAGE = {
-    'startedDateTime': 'date-time',
-    'request': {'method': '', 'url': '', 'headers': [], 'postData': 'body'},
-    'response': {'redirectURL': '', 'status': 200},
+    "startedDateTime": "date-time",
+    "request": {"method": "", "url": "", "headers": [], "postData": "body"},
+    "response": {"redirectURL": "", "status": 200},
 }
 
 # page with status_code=304
 PAGE_304 = {
-    'startedDateTime': 'date-time',
-    'request': {'method': '', 'url': '', 'headers': [], 'postData': 'body'},
-    'response': {'redirectURL': '', 'status': 304},
+    "startedDateTime": "date-time",
+    "request": {"method": "", "url": "", "headers": [], "postData": "body"},
+    "response": {"redirectURL": "", "status": 304},
 }
 
 # list for the page with status_code=302
 PAGES_200 = {
-    'log': {
-        'entries': [
+    "log": {
+        "entries": [
             # page with status code = 302 -> index 0
             {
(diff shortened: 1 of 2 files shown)


jb098:
Intial code looks good. I think it would be worth adding a unit test for this before we merge anything.

Sarah-J-S:
Indeed it would.

Sarah-J-S:
Well, it's a test. I'm not proud of it but it does what I think it needs to do. Please review with extreme prejudice.

test/test_har_to_mite.py
```diff
@@ -182,42 +183,58 @@ def test_set_headers():
...
+    if os.path.exists(testfile):
+        os.remove(testfile)
+    if os.path.exists(testfile10secs):
+        os.remove(testfile10secs)
+
+    har_convert_to_mite("test.har", "testharconvmite.txt", 0)
+    har_convert_to_mite("test.har", "testharconvmite10secs.txt", 10)
+
+    assert os.path.exists(testfile)
+    assert os.path.exists(testfile10secs)
```
aecay, line 223: I think we want some assertions on the new behavior that you have introduced -- i.e. do the delays in the generated code match those in the har file?

Sarah-J-S: I agree. While I couldn't see a way to do that without reams of code on Friday, I may have a quicker and dirtier method in mind after a weekend's reflection.

Sarah-J-S:
Extra testing added to ensure that the extracted wait times are what we expect to see.

test/test_har_to_mite.py
Sarah-J-S: Quick and dirty test now in place.

test/test_har_to_mite.py
```diff
@@ -182,42 +183,64 @@ def test_set_headers():
...
+
+    with open(testfile, "r") as testharoutfile:
+        testharStr = testharoutfile.read()
+    assert testharStr.find("await sleep(0)") != -1
```
aecay, line 227:
```suggestion
    assert "await sleep(0)" in testharStr
```

aecay (approved):
One suggested tweak to the test assertions but this lgtm

test/test_har_to_mite.py
aecay: Brilliant, thanks 🙂

test/test_har_to_mite.py
aecay: (and so on for the other assertions)

Sarah-J-S: That looks neater. I shall make it so. Another thing learned about Python.

test/test_har_to_mite.py
```diff
@@ -182,42 +183,64 @@ def test_set_headers():
...
+        os.remove(testfile10secs)
+
+    har_convert_to_mite("test.har", "testharconvmite.txt", 0)
+    har_convert_to_mite("test.har", "testharconvmite10secs.txt", 10)
+
+    assert os.path.exists(testfile)
+    assert os.path.exists(testfile10secs)
+
+    with open(testfile, "r") as testharoutfile:
+        testharStr = testharoutfile.read()
```
jb098, line 226: One very small comment that it's more pythonnic to use snake_case but I'll send you the tick anyway.

jb098 (approved):

test/test_har_to_mite.py
Sarah-J-S: My future code will strive for greater pythonicity. At least I've given up Hungarian notation.

PR #195, opened 2021-04-19
Add embedded resources option to browser decorator

cosmaprc:


mite_browser/__init__.py
```diff
@@ -41,11 +41,11 @@ def url_builder(base_url, *args, **kwargs):
     return url
 
 
-def browser_decorator(separation=0):
+def browser_decorator(separation=0, embedded_res=False):
```
aecay, line 44:
```suggestion
def browser_decorator(separation=0, embedded_resources=False):
```

(I know it was like this in the previous code, but it's a bad name so we should avoid perpetuating it...)

mite_browser/__init__.py
```diff
@@ -41,11 +41,11 @@ def url_builder(base_url, *args, **kwargs):
...
     def wrapper_factory(func):
         @mite_http
         async def wrapper(context, *args, **kwargs):
-            context.browser = Browser(context)
+            context.browser = Browser(context, embedded_res)
```
aecay, line 48:
```suggestion
            context.browser = Browser(context, embedded_resources)
```

mite_browser/__init__.py
```diff
@@ -41,11 +41,11 @@ def url_builder(base_url, *args, **kwargs):
     return url
 
 
-def browser_decorator(separation=0):
+def browser_decorator(separation=0, embedded_res=False):
     def wrapper_factory(func):
         @mite_http
```
aecay, line 46: While you are making changes, do you mind adding `wraps(func)` here (from functools)?  It helps with debugging in a few misc ways.  I've recently added it to other parts of mite (https://github.com/sky-uk/mite/pull/190) but I didn't do mite_browser because I (mistakenly, it turns out) thought it was in legacy-only mode.

aecay (approved):

mite_browser/__init__.py
cosmaprc: Done

Sarah-J-S (approved):

PR #201, opened 2021-05-14
Add SeleniumWrapper current_url fn

cosmaprc:


Matt2298 (approved):

mite_selenium/__init__.py
```diff
@@ -170,6 +170,8 @@ def _wait_for(self, condition_func, locator, timeout=5):
     def switch_to_iframe(self, locator):
         self._remote.switch_to.frame(self._remote.find_element(*locator))
 
+    def current_url(self):
+        return self._remote.current_url
```
aecay, line 174: Do you think we should make this an `@property` to match the selenium API?  I actually don;t mind one way or the other but figured I'd ask...

aecay (approved):

PR #202, opened 2021-05-18
Add selenium wait_for_url fn

cosmaprc:


Sarah-J-S (approved):

mite_selenium/__init__.py
```diff
@@ -159,13 +159,19 @@ def wait_for_elements(self, locator, timeout=5):
...
 
+    def wait_for_url(self, locator, timeout=5):
+        return self._wait_for(EC.url_to_be, locator, timeout)
+
     def _wait_for(self, condition_func, locator, timeout=5):
         try:
             return WebDriverWait(self._remote, timeout).until(condition_func(locator))
         except TimeoutException as te:
-            raise MiteError(
-                f"Timed out trying to find element '{locator}' in the dom"
```
jarnfield, line 167: It's not your introduction, but looks to me like this exception handling was initially incorrectly placed, and the change has only made things a little better. This error handler is making assumptions about the condition_func. You've addressed that for one case, but when considering this function in isolation; "timed out trying to find element" is still a default case for things that aren't of type `url_to_be`, which is too presumptous.

I think it would be better to handle these exceptions in the callers? One less bit of branching and no need to have to mentally clock that this function also needs maintainance when adding new wait_for functions in the future.

jarnfield: In fact, I can see how initially it may have arguably made sense, but now it probably makes more sense to move the try: catch:

jarnfield: Also - sorry. You know what they say, give a developer 1 line to review and they'll find 100 problems with it.. or something.

jarnfield (approved):

PR #203, opened 2021-05-24
Add selenium-wire support

cosmaprc:
Use selenium remote driver quit instead of stop to avoid exhausting the selenium browser sessions/instances while waiting for browser timeouts

### DIFF
--- a/mite_selenium/__init__.py
+++ b/mite_selenium/__init__.py
@@ -28,10 +28,10 @@ class _SeleniumWrapper:
             "webdriver_command_executor", "http://127.0.0.1:4444/wd/hub"
         )
         self._keep_alive = self._context.config.get("webdriver_keep_alive", False)
-        self._file_detector = self._spec_import_if_none("webdriver_file_detector")
-        self._proxy = self._spec_import_if_none("webdriver_proxy")
-        self._browser_profile = self._spec_import_if_none("webdriver_browser_profile")
-        self._options = self._spec_import_if_none("webdriver_options")
+        self._file_detector = self._spec_import_if_not_none("webdriver_file_detector")
+        self._proxy = self._spec_import_if_not_none("webdriver_proxy")
+        self._browser_profile = self._spec_import_if_not_none("webdriver_browser_profile")
+        self._options = self._spec_import_if_not_none("webdriver_options")
 
         # Required param
         self._capabilities = self._context.config.get("webdriver_capabilities")
--- a/mite_selenium/__init__.py
+++ b/mite_selenium/__init__.py
@@ -1,12 +1,14 @@
 import logging
+import socket
 import time
 from contextlib import asynccontextmanager
 from functools import wraps
 
 from selenium.common.exceptions import TimeoutException
-from selenium.webdriver import Remote
+from selenium.webdriver import Remote as SeleniumRemote
 from selenium.webdriver.support import expected_conditions as EC
 from selenium.webdriver.support.ui import WebDriverWait
+from seleniumwire.webdriver import Remote as SeleniumWireRemote
 
 from mite.exceptions import MiteError
 from mite.utils import spec_import
@@ -34,6 +36,8 @@ class _SeleniumWrapper:
         # Required param
         self._capabilities = self._context.config.get("webdriver_capabilities")
         self._capabilities = spec_import(self._capabilities)
+        self._seleniumwire_options = self._spec_import_if_none("seleniumwire_options")
+        self._is_selenium_wire = self._seleniumwire_options is not None
 
     def _spec_import_if_none(self, config_option):
         value = self._context.config.get(config_option, None)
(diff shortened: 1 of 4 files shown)


mite_selenium/__init__.py
```diff
@@ -43,19 +47,41 @@ def _spec_import_if_none(self, config_option):
...
-            desired_capabilities=self._capabilities,
-            command_executor=self._command_executor,
-            browser_profile=self._browser_profile,
-            proxy=self._proxy,
-            keep_alive=self._keep_alive,
-            file_detector=self._file_detector,
-            options=self._options,
-        )
+        try:
+            if self._is_selenium_wire:
```
aecay, line 51: Is there a usecase for not having selenium wire?  If not, we should just use it unconditionally.  The thing I can most reasily imagine is if selenium-wire is much slower than selenium itself (medium-bad) and if it's not possible to measure and subtract out that extra slowness from our metrics (very bad indeed).

mite_selenium/__init__.py
```diff
@@ -186,10 +212,18 @@ def _wait_for(self, condition_func, locator, err_msg, timeout=5):
     def switch_to_iframe(self, locator):
         self._remote.switch_to.frame(self._remote.find_element(*locator))
 
+    def add_request_interceptor(self, request_interceptor):
+        if self.is_selenium_wire:
+            self._remote.request_interceptor = request_interceptor
```
aecay, line 217: `else: raise Exception("request interceptor not supported")`

(but see above about using selenium-wire unconditionally, which makes this comment go away)

mite_selenium/__init__.py
```diff
@@ -43,19 +47,41 @@ def _spec_import_if_none(self, config_option):
...
+                self._remote = SeleniumRemote(
+                    desired_capabilities=self._capabilities,
+                    command_executor=self._command_executor,
+                    browser_profile=self._browser_profile,
+                    proxy=self._proxy,
+                    keep_alive=self._keep_alive,
+                    file_detector=self._file_detector,
+                    options=self._options,
+                )
+        except Exception as e:
```
aecay, line 77: This catch-all exception handling is a bit strange to see -- what kinds of exceptions do we expect to see?  Why doesn't the vanilla exception handler print them out correctly?

mite_selenium/__init__.py
jb098: I don't know much about selenium wire, but I do have the general principle that we should not assume what the end user of mite is going to want. I think if there were any side effect, it would make this decision for us.

mite_selenium/__init__.py
```diff
@@ -1,12 +1,14 @@
...
 import time
 from contextlib import asynccontextmanager
 from functools import wraps
 
 from selenium.common.exceptions import TimeoutException
-from selenium.webdriver import Remote
+from selenium.webdriver import Remote as SeleniumRemote
 from selenium.webdriver.support import expected_conditions as EC
 from selenium.webdriver.support.ui import WebDriverWait
+from seleniumwire.webdriver import Remote as SeleniumWireRemote
```
jb098, line 11: It looked like in other places that you were checking if selenium wire existed, is this not going to break if it doesn't?

cosmaprc: selenium-wire is now a requirement in the dependencies in setup.cfg so It should always get installed as a dependency and always be there for import. My checks were to see if the remote object was to be a selenium or selenium-wire instance.

mite_selenium/__init__.py
cosmaprc: In the selenium-wire docs it states 
> Selenium Wire has limited support for using the remote webdriver client.
So a user might need some functionality that is not available yet. Also like Jordan said about not assuming what the end user needs from mite.
I also think it's too early for us to tell if selenium-wire is much slower that selenium itself and if it's possible or not to measure and subtract that out of our metrics and so I would keep both for now.

aecay: Yeah, I agree with what Jordan is saying about avoiding side effects.  And the docs Cosmin quoted show that there are potential side effects (missing functionality).  So we need to support both.  In that case, I think:
1. the selenium wire specific code should be in a subclass of `_SeleniumWrapper` rather than in conditionals
2. the `mite_selenium` decorator should gain an `wire=True/False` argument that controls whether it returns a `_SeleniumWrapper` or a `_SeleniumWireWrapper`
3. (not strictly necessary but nice to have) we should make selenium wire an [optional dependency](https://setuptools.readthedocs.io/en/latest/userguide/dependency_management.html#optional-dependencies) so that it's not pulled in unless the user specifically requests it (obviously, they can't use the wire functionality unless the dependency is present -- we should detect that case in the `mite_selenium` decorator and raise an error)

aecay:
suggestions in https://github.com/sky-uk/mite/pull/204

jb098 (approved):

PR #205, opened 2021-05-26
Finagle load injection

aecay:
Create infrastructure for finagle load injection in mite.

See the README for full details.

Unrelated changes included in the PR:
- upgrade isort, with consequential changes to the isort config file
- remove the old cat_msgpack.py helper; we have `mite cat` now
- reformat files
  - `mite/__init__.py`
  - `mite/controller.py`
  - `mite/recorder.py`
  - `mite_amqp/__init__.py`
  - `test/perf/http_server.py`
  - `test/test_recorder.py`
  - `test/test_runner_tracker.py`
  - `test/test_volume_model.py`
  - `test/test_work_tracker.py`
  - `test-requirements.txt`
- fix the `--message-processors` argument to `mite scenario test` for the case when it's empty


Matt2298 (approved):

jarnfield (commented):
Can't wait to try it. A couple of breakpoints() that look like they might be remnants of debugging that might be worth pointing out in case you meant to remove them, but most of that kind of stuff will get ironed out when we use this functionality in anger 👍

jarnfield (approved):

aecay:
@johnarnfield I had a look at the breakpoints (of which I spotted two) -- both are in branches that (according to me) should never be entered, so if we do wind up there I think it's probably serious enough for us to get dropped into the debugger.  I'm slightly violating my own principles in terms of merging breakpoints into the public mite code, but I also think the best place for this code to live is on the master branch (not in a perma-open PR or side branch) until someone on the squad picks it up "for real", so I think I'm going to hold my nose and pragmatically press merge... 😕

PR #207, opened 2021-05-27
Allow parallel async transactions

jarnfield:


test/test_context.py
```diff
@@ -141,3 +142,37 @@ async def test_mite_error_backtrace_points_to_exn_source():
...
+        async with context.transaction("B"):
+            assert context.transaction_name == "A :: B"
+            async with context.transaction("C"):
+                assert context.transaction_name == "A :: B :: C"
+            assert context.transaction_name == "A :: B"
+        assert context.transaction_name == "A"
+
+
+@pytest.mark.asyncio
+async def test_parallel_transaction_names():
```
jarnfield, line 163: This is the test case that fails with the existing code

test/test_context.py
```diff
@@ -141,3 +142,37 @@ async def test_mite_error_backtrace_points_to_exn_source():
...
+        async with context.transaction(label):
+            assert context.transaction_name == label
+            await asyncio.sleep(1)
+            assert context.transaction_name == label
+
+    await asyncio.gather(
+        sleepy_transaction("A"),
+        sleepy_transaction("B"),
+        sleepy_transaction("C"),
+        sleepy_transaction("D"),
```
aecay, line 177: Two transactions are sufficient to make it fail, right?  We should have the minimal test case...

mite/context.py
```diff
@@ -53,25 +54,37 @@ def should_stop(self):
             return self._should_stop_func()
         return False
 
+    @property
+    def transaction_name(self):
+        return active_transaction.get()[0]
+
+    @property
+    def transaction_id(self):
```
aecay, line 62: I think this wants to have an underscore-prefixed name since it's an implementation detail that users of mite shouldn't care about

mite/context.py
```diff
@@ -53,25 +54,37 @@ def should_stop(self):
...
+        return active_transaction.get()[1]
+
+    def extend_transaction(self, name):
```
aecay, line 65: Same: I think this should be prefixed with an underscore and (thus) not be part of the public api of Context

mite/context.py
```diff
@@ -4,6 +4,7 @@
 import sys
 import time
 import traceback
+from contextvars import ContextVar
```
aecay, line 7: ...not...in...alphabetical...order... 😅 

obsessive

aecay (approved):
I understand the problem now.  This looks like the correct solution.  Couple very small suggestions...

PR #210, opened 2021-07-24
suggestions for acurl rewrite

aecay:
- add some documentation in comments around some of the more obscure areas of the code
  - while reviewing/documenting these, some of them showed up as apparently not needed any longer which is (tentative) good news!
- misc small fixes
- line coverage for cython -- it took a fair amount of hacks to enable for the test suite only, and the hacks work -- but the tests crash hard (SIGSEGV) when it's actually enabled :cry: 
- fix the Request.headers property (and associated tests) -- it was confused about bytes vs str


Matt2298 (approved):
LGTM, I'll take a look at the remaining comments on the other PR once I'm back from Lithuania 👍

DavAnnunz (approved):

PR #213, opened 2021-08-14
Acurl cleanups and memleak fixes

aecay:
Here are the fixes and things we talked about a couple weeks ago.  I've split the commits up by purpose, and also ordered them into blocks for ease of reviewing/digesting.

- Commits up to 920d (inclusive) are docs/tooling/cosmetic
- f2be is a fix to the mite test suite removing a questionable mock that would cause the new acurl code to trip
- 73bf and 665c are documentation fixes inside the acurl code
- 30c0 and 7c18 are code changes unrelated to mem leak stuff
- and finally, 7d57 to the end are mem leak fixes

### DIFF
--- a/Dockerfile.perftest
+++ b/Dockerfile.perftest
@@ -1,4 +1,4 @@
-FROM python:3.7
+FROM python:3.9
 
 COPY ./test/perf/test-requirements.txt /test-requirements.txt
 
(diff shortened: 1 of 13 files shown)


Matt2298 (approved):
Looking really good now, thanks for including some valgrind documentation too 👍

cosmaprc (approved):

PR #214, opened 2021-09-22
Flush the raw message after writing

ryanlinnit-sky:
Occasionally when sending a message it would not write to the collector file. The buffered writer is only written to under the conditions [[0](https://docs.python.org/3/library/io.html#io.BufferedWriter)]: 
 - when the buffer gets too small for all pending data;
 - when flush() is called;
 - when a seek() is requested (for BufferedRandom objects);
 - when the BufferedWriter object is closed or destroyed.

The odd message would get missed, I believe due to the `Collector` object being destroyed before a `close` is called on our writer object. This PR ensures a `flush` is called after every write.

### DIFF
--- a/mite/cli/test.py
+++ b/mite/cli/test.py
@@ -153,6 +153,10 @@ def test_scenarios(test_name, opts, scenarios, config_manager):
     # Run one last report before exiting
     controller.report(receiver.recieve)
     has_error = http_stats_output is not None and http_stats_output.error_total > 0
+
+    # Ensure any open files get closed
+    del receiver._raw_listeners
+
     sys.exit(int(has_error))
 
 
--- a/mite/cli/test.py
+++ b/mite/cli/test.py
@@ -156,6 +156,7 @@ def test_scenarios(test_name, opts, scenarios, config_manager):
 
     # Ensure any open files get closed
     del receiver._raw_listeners
+    del receiver._listeners
 
     sys.exit(int(has_error))
 
--- a/mite/collector.py
+++ b/mite/collector.py
@@ -32,6 +32,9 @@ class Collector:
         self._file_counter = count()
         self._rotate_current_file()
 
+    def __del__(self):
+        self._current.close()
+
     @property
     def _current_fn(self):
         return os.path.join(self._target_dir, "current")
@@ -53,7 +56,6 @@ class Collector:
             self._current.write(json.dumps(decoded).encode() + b"\n")
         else:
             self._current.write(msg)
-        self._current.flush()
 
     def _rotate_current_file(self):
         self._msg_count = 0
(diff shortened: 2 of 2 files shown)


aecay:
Calling flush after every write is potentially going to be quite slow.  What about using a finalizer (`__del__`) on the Collector class to ensure that the most recent `_current` file is `close()`d when the instance gets deallocated?  Annoyingly:

> It is not guaranteed that `__del__()` methods are called for objects that still exist when the interpreter exits. ([link](https://docs.python.org/3/reference/datamodel.html#object.__del__))

So we might have to use the [atexit](https://docs.python.org/3/library/atexit.html) library too, depending on whether the collector remains live at exit...  (Or we could rearrange the code so that the collector goes out of scope sooner.)

ryanlinnit-sky:
> 
> 
> Calling flush after every write is potentially going to be quite slow. What about using a finalizer (`__del__`) on the Collector class to ensure that the most recent `_current` file is `close()`d when the instance gets deallocated? Annoyingly:
> 
> > It is not guaranteed that `__del__()` methods are called for objects that still exist when the interpreter exits. ([link](https://docs.python.org/3/reference/datamodel.html#object.__del__))
> 
> So we might have to use the [atexit](https://docs.python.org/3/library/atexit.html) library too, depending on whether the collector remains live at exit... (Or we could rearrange the code so that the collector goes out of scope sooner.)

I was concerned about the impact it may have and I did try adding
```
def __del__(self):
    self._current.close()
```

But it seems it doesn't get called. I'll have a look at atexit. Thanks

ryanlinnit-sky:
@aecay but I can specifcally call `del receiver._raw_listeners` at the end of the test, which does call our `__del__` method. :D

mite/cli/test.py
```diff
@@ -153,6 +153,10 @@ def test_scenarios(test_name, opts, scenarios, config_manager):
     # Run one last report before exiting
     controller.report(receiver.recieve)
     has_error = http_stats_output is not None and http_stats_output.error_total > 0
+
+    # Ensure any open files get closed
+    del receiver._raw_listeners
```
aecay, line 158:
```suggestion
    del receiver
```

I think it should be like this, to more robustly ensure all resources are cleaned up (the alternative would be to also `del receiver._listeners`)

aecay (commented):
Nice solution!  Just one comment, but overall this approach LGTM

mite/cli/test.py
ryanlinnit-sky: With just `del receiver` it never cleans up the Collector, so I just did your alternative

aecay: hmm, odd.  I wonder why :thinking: although it doesn't matter _that_ much.

aecay (approved):
lgtm!

Sarah-J-S (approved):

cosmaprc (approved):

Matt2298 (approved):

PR #216, opened 2021-10-20
set version of selenium which does not require additional dependencies

skysquirrel2021:
the changes have been built, but it is probably worthwhile someone else building the changes to ensure I have not done anything stupid

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -12,7 +12,7 @@ install_requires =
     msgpack
     nanomsg
     pyzmq
-    selenium
+    selenium==3.141.0
     uvloop
     websockets
 setup_requires = pytest-runner
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,6 +20,9 @@ ADD . / /mite/
 WORKDIR /mite/acurl
 RUN pip install --no-cache-dir -e .
 
+WORKDIR /mite/acurl_ng
+RUN pip install --no-cache-dir -e .
+
 WORKDIR /mite
 RUN pip install --no-cache-dir -e .
 


setup.cfg
```diff
@@ -12,7 +12,7 @@ install_requires =
     msgpack
     nanomsg
     pyzmq
-    selenium
+    selenium==3.141.0
```
aecay, line 15: Rather than pinning this to a specific version, it might be better to use a range like `selenium<4`.  I am assuming that the new (last week) version 4 is what breaks things.  By pinning to a range, we ensure that we'd continue to pull in updates to the 3.X release train in case of future bugfixes or security patches.  (I don't know how likely there are to be any such changes in the future -- possibly unlikely.  But we don't need to rule them out from the get-go.)  WDYT?

I'd also be interested to hear what the problems have been in moving to a newer selenium-python version, and getting the changes needed to migrate onto the mite roadmap (such as it is) -- but that's very much a "nice to have" and not a requirement to merge the PR.

skysquirrel2021: I agree entirely that pinning a version is probably not the best idea! However, it looks like anything above 4.0.0a5 drags in a whole bunch of dependencies (which includes the need to install the rust compiler, which in itself seems to create issues), so, on balance, I suspect, that for the near future at least, it is probably better to  have a working mite build which has a pinned selenium version, rather than have a broken build. I also think that further investigation as to what benefits the later selenium versions have would be worthwhile.
The version I have pinned to, is the last known good "release" version. Whereas, all the 4x versions, up to HEAD, are pre-release / alpha, so, I am a bit reticent to use those versions in prod. Counter thoughts?

skysquirrel2021:
if anyone has overview of mite-ci and why is may be failing, that would be good!

aecay:
> if anyone has overview of mite-ci and why is may be failing

```
[2021-10-21T06:27:00.628Z] Traceback (most recent call last):
[2021-10-21T06:27:00.628Z]   File "/home/jenkins/home/workspace/identity/mite/mite-ci/mite/.tox/py39/lib/python3.9/site-packages/pip/_internal/cli/base_command.py", line 173, in _main
[2021-10-21T06:27:00.628Z]     status = self.run(options, args)
[2021-10-21T06:27:00.628Z]   File "/home/jenkins/home/workspace/identity/mite/mite-ci/mite/.tox/py39/lib/python3.9/site-packages/pip/_internal/cli/req_command.py", line 203, in wrapper
[2021-10-21T06:27:00.628Z]     return func(self, options, args)
[2021-10-21T06:27:00.628Z]   File "/home/jenkins/home/workspace/identity/mite/mite-ci/mite/.tox/py39/lib/python3.9/site-packages/pip/_internal/commands/install.py", line 315, in run
[2021-10-21T06:27:00.628Z]     requirement_set = resolver.resolve(
[2021-10-21T06:27:00.628Z]   File "/home/jenkins/home/workspace/identity/mite/mite-ci/mite/.tox/py39/lib/python3.9/site-packages/pip/_internal/resolution/resolvelib/resolver.py", line 94, in resolve
[2021-10-21T06:27:00.628Z]     result = self._result = resolver.resolve(
[2021-10-21T06:27:00.629Z]   File "/home/jenkins/home/workspace/identity/mite/mite-ci/mite/.tox/py39/lib/python3.9/site-packages/pip/_vendor/resolvelib/resolvers.py", line 472, in resolve
[2021-10-21T06:27:00.629Z]     state = resolution.resolve(requirements, max_rounds=max_rounds)
[2021-10-21T06:27:00.629Z]   File "/home/jenkins/home/workspace/identity/mite/mite-ci/mite/.tox/py39/lib/python3.9/site-packages/pip/_vendor/resolvelib/resolvers.py", line 383, in resolve
[2021-10-21T06:27:00.629Z]     raise ResolutionTooDeep(max_rounds)
[2021-10-21T06:27:00.629Z] pip._vendor.resolvelib.resolvers.ResolutionTooDeep: 2000000
```

:nauseated_face: looks like a problem with the new-ish pip dependency resolver, somehow.  The new set of package versions is having difficulty being reduced to a specific set of versions.  I know that it occasionally requires obscure rituals to pacify this piece of code, but I've never encountered it in the wild before...

skysquirrel2021:
ok, hopefully it will resolve itself on the next build, otherwise, who do i need to prod to help get jenkins to play :)

PR #221, opened 2021-11-09
Add a new option to set an error threshold in the scenario command

DavAnnunz:
Sometimes when running a big amount of journeys a single error can be irrelevant and making fail the whole job can be annoying. This will allow mite to set an exit status of 0 even if we got some error if the number will be lower than the set threshold.

### DIFF
--- a/mite/__main__.py
+++ b/mite/__main__.py
@@ -67,6 +67,7 @@ Options:
     --message-processors=PROCESSORS         Classes to connect to the message bus for local testing [default: mite.logoutput:HttpStatsOutput,mite.logoutput:MsgOutput]
     --prettify-timestamps                   Reformat unix timestamps to human readable dates
     --journey-logging                       Log errors on a per journey basis
+    --max-errors-threshold=THRESHOLD        Set the maximum amount of error accepted before set exit stauts to 1 [default: 0]
 """
 import asyncio
 import logging
--- a/mite/__main__.py
+++ b/mite/__main__.py
@@ -67,7 +67,7 @@ Options:
     --message-processors=PROCESSORS         Classes to connect to the message bus for local testing [default: mite.logoutput:HttpStatsOutput,mite.logoutput:MsgOutput]
     --prettify-timestamps                   Reformat unix timestamps to human readable dates
     --journey-logging                       Log errors on a per journey basis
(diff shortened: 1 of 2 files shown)


mite/__main__.py
```diff
@@ -67,6 +67,7 @@
     --message-processors=PROCESSORS         Classes to connect to the message bus for local testing [default: mite.logoutput:HttpStatsOutput,mite.logoutput:MsgOutput]
     --prettify-timestamps                   Reformat unix timestamps to human readable dates
     --journey-logging                       Log errors on a per journey basis
+    --max-errors-threshold=THRESHOLD        Set the maximum amount of error accepted before set exit stauts to 1 [default: 0]
```
Matt2298, line 70:
```suggestion
    --max-errors-threshold=THRESHOLD        Set the maximum amount of error accepted before setting exit status to 1 [default: 0]
```

Matt2298 (approved):
LGTM after typos

mite/__main__.py
aecay: I would go with "number of errors" rather than "amount of error"

mite/cli/test.py
```diff
@@ -152,7 +152,9 @@ def test_scenarios(test_name, opts, scenarios, config_manager):
     loop.run_until_complete(asyncio.wait(coroutines, return_when=asyncio.FIRST_COMPLETED))
     # Run one last report before exiting
     controller.report(receiver.recieve)
-    has_error = http_stats_output is not None and http_stats_output.error_total > 0
+    has_error = http_stats_output is not None and http_stats_output.error_total > int(
+        opts.get("--max-errors-threshold")
+    )
```
aecay, line 157: Would it make sense to detect the case where 1 > `opts.get("--max-errors-threshold")` >= 0 and, in that case, treat the value as a percentage of total requests instead?  That's an additional feature on top of what you implemented (so it's not a requirement to merge the PR), but it seems like it would be useful...

aecay (approved):
a couple small comments but overall lgtm

mite/cli/test.py
DavAnnunz: I had a look if we already have a variable for the total amount of journeys ran but looks it is not there. I like the idea though, I'll write it down as an improvement to add

PR #224, opened 2021-11-12
Raise AcurlError on non-zero exit code

cosmaprc:
Curl will return a non-zero exit code when something goes wrong.This change is to raise an AcurlError that describes the error code. See https://curl.se/libcurl/c/libcurl-errors.html and https://everything.curl.dev/usingcurl/returns for details.

Skipping http 100 tests for now since this change revealed a problem with them a.i. they hang because curl returns CURLE_RECV_ERROR (56) instead of CURLE_OK (0).

### DIFF
--- a/acurl_ng/src/acurl.pyx
+++ b/acurl_ng/src/acurl.pyx
@@ -11,6 +11,9 @@ include "request.pyx"
 include "response.pyx"
 include "session.pyx"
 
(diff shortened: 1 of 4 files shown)


Sarah-J-S (approved):
Neat!

aecay (approved):
one small suggestion but this code lgtm :grinning:

PR #225, opened 2021-11-17
Added headless flag to SeleniumWrapper class

richardfeather1:
Adding attribute to pass headless setting through config IPB-14599

### DIFF
--- a/mite_selenium/__init__.py
+++ b/mite_selenium/__init__.py
@@ -32,6 +32,7 @@ class _SeleniumWrapper:
         self._proxy = self._spec_import_if_not_none("webdriver_proxy")
         self._browser_profile = self._spec_import_if_not_none("webdriver_browser_profile")
         self._options = self._spec_import_if_not_none("webdriver_options")
+        self._options.headless = self._context.config.get("webdriver_options_headless")
 
         # Required param
         self._capabilities = self._context.config.get("webdriver_capabilities")
--- a/mite_selenium/__init__.py
+++ b/mite_selenium/__init__.py
@@ -16,6 +16,8 @@ from mite.utils import spec_import
 
 logger = logging.getLogger(__name__)
 
+print(type(Options()))
+
 
 class _SeleniumWrapper:
     def __init__(self, context):
@@ -33,7 +35,9 @@ class _SeleniumWrapper:
         self._proxy = self._spec_import_if_not_none("webdriver_proxy")
         self._browser_profile = self._spec_import_if_not_none("webdriver_browser_profile")
         self._options = self._spec_import_if_not_none("webdriver_options")
-        if self._options:
+        if not self._options:
+            self._options = Options()
+        if type(self._options) != bool:
             self._options.headless = self._context.config.get("webdriver_options_headless", True)
         # Required param
         self._capabilities = self._context.config.get("webdriver_capabilities")
--- a/mite_selenium/__init__.py
+++ b/mite_selenium/__init__.py
@@ -16,8 +16,6 @@ from mite.utils import spec_import
 
 logger = logging.getLogger(__name__)
 
-print(type(Options()))
-
 
 class _SeleniumWrapper:
     def __init__(self, context):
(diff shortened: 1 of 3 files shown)


mite_selenium/__init__.py
```diff
@@ -32,6 +32,7 @@ def __init__(self, context):
         self._proxy = self._spec_import_if_not_none("webdriver_proxy")
         self._browser_profile = self._spec_import_if_not_none("webdriver_browser_profile")
         self._options = self._spec_import_if_not_none("webdriver_options")
+        self._options.headless = self._context.config.get("webdriver_options_headless", True)
```
aecay, line 35: 1. If `self._options` is None then this will blow up when it tries to do `None.headless = True`
2. If `self._options` is not None, then you should have set `headless` on that object in the first place

There's still a need for some kind of default, but I think it needs to look like:

```
self._options = self._spec_import_if_not_none("webdriver_options")
if self._options is None:
    self._options = SeleniumOptions(headless=True)
```

We need to import SeleniumOptions.  A quick glance at the selenium API makes it look like there are separate options classes for chrome and FF; idk if there's a base class we can use in both contexts or whether we need to specialize (and I also don't know and can't remember if we're actually using chrome or FF, or leaving it to a user choice...  I assume you'll be able to sort all of that out...)

mite_selenium/__init__.py
```diff
@@ -32,8 +33,8 @@ def __init__(self, context):
         self._proxy = self._spec_import_if_not_none("webdriver_proxy")
         self._browser_profile = self._spec_import_if_not_none("webdriver_browser_profile")
         self._options = self._spec_import_if_not_none("webdriver_options")
-        self._options.headless = self._context.config.get("webdriver_options_headless", True)
-
+        if self._options:
+            self._options.headless = self._context.config.get("webdriver_options_headless", True)
```
aecay, line 37: If `self._options` has a value, that's because it was set to something by the user.  Since the user has control of the object whose value is assigned to `self._options`, the user can just set headless on that object.  So we don't need a second user option (`webdriver_options_headless`) for that case.

OTOH, I do see the value of having `webdriver_options_headless` in the case where the user is not setting the options to anything.  It would be convenient not to need to create a whole `Options` object (imported from selenium lib) just to set headless, but rather be able to set it via a simple boolean.  We don't _have_ to offer this flexibility, but we could.

If you accept this analysis, then the conditional here (`if self._options`) has the wrong polarity (we should only be paying attention to `webdriver_options_headless` in the case `if not self.options`)

mite_selenium/__init__.py
```diff
@@ -9,6 +9,7 @@
 from selenium.webdriver import Remote as SeleniumRemote
 from selenium.webdriver.support import expected_conditions as EC
 from selenium.webdriver.support.ui import WebDriverWait
+from selenium.webdriver.chrome.options import Options
```
richardfeather1, line 12: Need to find a way to establish is user is using Chrome or Firefox to pull the right Options object

mite_selenium/__init__.py
```diff
@@ -32,7 +33,10 @@ def __init__(self, context):
         self._proxy = self._spec_import_if_not_none("webdriver_proxy")
         self._browser_profile = self._spec_import_if_not_none("webdriver_browser_profile")
         self._options = self._spec_import_if_not_none("webdriver_options")
-
+        if not self._options:
+            self._options = Options()
+        if type(self._options) != bool:
+            self._options.headless = self._context.config.get("webdriver_options_headless", True)
```
richardfeather1, line 39: Doesn't feel right.... but the test is failing as we use a boolean of True, so we need something to get past this point.

test/test_webdriver.py
```diff
@@ -51,7 +52,7 @@ def test_config_defaults():
     assert wrapper._file_detector is None
     assert wrapper._proxy is None
     assert wrapper._browser_profile is None
-    assert wrapper._options is None
+    assert wrapper._options is not None
```
richardfeather1, line 55: We now should always pass an Options object...whether we're right to, I'm not 100% certain!

PR #226, opened 2021-11-24
Fix HTTP 100 test

ryanlinnit-sky:
Fixes the broken tests from #224

## `test_response_headers_with_HTTP_100` test

Instead of creating a dummy http server, it's using pytest-httpserver and sending the `Expect: 100-continue` header.

We can confirm that we get a "100 Continue" followed by "200 OK" by looking at `r.header`:

```HTTP/1.1 100 Continue\r\n\r\nHTTP/1.0 200 OK\r\nFoo: bar\r\nContent-Type: text/plain; charset=utf-8\r\nContent-Length: 0\r\nServer: Werkzeug/2.0.1 Python/3.9.9\r\nDate: Wed, 24 Nov 2021 11:40:31 GMT\r\n\r\n```

## `test_response_headers_with_multiple_HTTP_100` test

Removed this test. `CURLINFO_RESPONSE_CODE` should always have the last received response code, so we can be pretty confident we're safe against this corner case.

### DIFF
--- a/acurl_ng/tests/test_response_ng.py
+++ b/acurl_ng/tests/test_response_ng.py
@@ -41,7 +41,9 @@ async def test_response_headers_with_HTTP_100(httpserver, acurl_session_ng):
     httpserver.expect_request("/foo").respond_with_response(
         Response(response="", status=200, headers=hdrs)
     )
-    r = await acurl_session_ng.get(httpserver.url_for("/foo"), headers={"Expect": "100-continue"})
+    r = await acurl_session_ng.get(
+        httpserver.url_for("/foo"), headers={"Expect": "100-continue"}
+    )
 
     assert "Foo" in r.headers
     assert r.headers["Foo"] == "bar"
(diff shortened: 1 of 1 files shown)


aecay (approved):
nice one!

ryanlinnit-sky:
> nice one!

Still a draft at the minute. Trying to figure out how to resolve the `test_response_headers_with_multiple_HTTP_100` test. Can't find a simple way to send multiple `100 CONTINUE`. I may just find justification for just getting rid of `test_response_headers_with_multiple_HTTP_100`..

aecay:
> just getting rid of test_response_headers_with_multiple_HTTP_100

that seems sensible to me.  It was a corner case and we were never sure whether it could happen in the wild (though I think the RFCs allow it).  We covered it just because we could, not out of any particular motivated need.  I personally would not lose any sleep over that test getting ditched.

cosmaprc (approved):

PR #227, opened 2021-12-29
Add paint metrics to selenium

jb098:


mite_selenium/__init__.py
```diff
@@ -85,10 +85,20 @@ def _send_page_load_metrics(self):
...
+                "return performance.getEntriesByType('paint')"
+            )
+
+            timings = self._extract_entries(performance_entries)[0]
             if timings is None:
                 return
 
+            _paint_timings = self._extract_entries(paint_entries, expected=2)
+            if _paint_timings is None:
+                return
```
aecay, line 98: This code won't send any metrics if we have access to navigation metrics but not paint metrics (for whatever reason).  It might make sense to structure the code like this:

```
get_navigation_metrics()
metrics = { ... }
paint_metrics = get_paint_metrics()
if paint_metrics is not None:
    metrics["first_contentful_paint"] = ...
    ...
return metrics
```

So that we still get navigation metrics if paint metrics are missing.  But maybe it's not worth bothering, if the paint metrics will always be there....

aecay (approved):
Seems very sensible -- I had one question about what might happen when the paint metrics are missing and whether we still need to report navigation metrics in that case -- but IDK if that situation will ever actually happen / if it is worth coding defensively for.

mite_selenium/__init__.py
jb098: Great point Aaron. Got taken out with covid last week so parked this but I'll be jumping back on it today and I'll fix this.

Matt2298 (approved):

cosmaprc (approved):

PR #231, opened 2022-02-15
IPB-16472 Fix Jenkins Slave Image Rust issue

Sarah-J-S:
Using the latest version of Selenium causes incompatibility issues with Rust. Setting the version to less than 4 prevents this issue. 

Note that this will need to be revised when we upgrade to Selenium 4. The card for that work - https://agile.at.sky/browse/IPB-15616 - has been updated to capture this.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,6 +1,8 @@
 [metadata]
 name = mite
 
+# TODO - Update (or remove) the Selenium version when migrating to v4. The version is explicitly defined
+#        as less than 4 as here are backward compatibility issues with Rust from version 4 on
 [options]
 install_requires =
     acurl
@@ -11,7 +13,7 @@ install_requires =
     msgpack
     nanomsg
     pyzmq
-    selenium
+    selenium<4
     uvloop
     websockets
 setup_requires = pytest-runner


skysquirrel2021 (approved):

setup.cfg
```diff
@@ -11,7 +13,7 @@ install_requires =
     msgpack
     nanomsg
     pyzmq
-    selenium
+    selenium<4
```
aecay, line 16: I thought we solved this with https://github.com/sky-uk/mite/pull/216 (as an interim solution) and then https://github.com/sky-uk/mite/pull/217 (which was supposed to eliminate the need for a rust build chain in the mite docker image).

It would be good to understand what has broken about that solution...my suspicion (without having looked very hard, or indeed at all yet) is that the cryptography library that's getting pulled in by #217 might have gotten outdated, so now pip install is trying to pull in a newer version (which in turn is depending on a rust build env).

It might be better to pin our version of cryptography to prevent this from happening.  Other options could be:
- to pin to a specific version of selenium 4 that is compatible with our baked-in cryptography lib or
- to upgrade our underlying alpine image(s) to hopefully gain access to a newer pre-built cryptography lib.

Sarah-J-S: I was hoping that you would see this PR. I looked at those PRs but unfortunately the rust installation is broken again. As rust was the symptom that was where I looked first. Migrating to Selenium 4 is a non-trivial exercise as there some incompatibilities that we would need to address (as per IPB-15616 above). I'll taker a closer look at the version of the cryptography library that is being installed and see i there's a better fix.

aecay: :+1: I thought we were using selenium 4 already, after those PRs.  But a version bump of selenium and/or pycryptography must have broken it :disappointed: 

If things are broken for you/the team, we can certainly merge this in -- but I want to make sure we don't lose sight of the goal to solve this systematically, which is proving more elusive than it seemed just a few short days ago....

Sarah-J-S: Cool, thank you. I'm leaving notes where I can to indicate that we must Find A Better Way, but probably after adopting Selenium 4.

PR #232, opened 2022-02-16
add patch functionality

miguel8998:
## Description 
Added methods for mite to use PATCH to make calls 

## testing 
created these 2 journeys to create some data in my server and then update the data with a patch

image

Having a look at the data before and after the patch we can see that it changes as expected 

### Before PATCH
image
### After PATCH
image


jb098 (approved):
Looks good to me Miguel!

AaliqMoin (approved):

PR #240, opened 2022-06-21
cleaning up mite

ryanlinnit-sky:
Mite code cleanup, trying to bring a little bit of consistency to the code. Follow peps where ideal.  

- Update python versions to 3.10
 - Merge assignment followed by conditional check using a named expression.
 Cleaner code - https://peps.python.org/pep-0572/
 - Raise a specific error, instead of just `Exception`
 - Replace Interpolation With Fstring
 - Converts `for` loop into a list comprehension
 - Create the list with values instead of creating an empty list and appending to it
 - Replace a for append loop with list extend
 - Don't assign to builtin variables, such as sum.
 - Returning a result directly instead of assigning a variable then returning that
 - Simplifies conditional logic using De Morgan's laws
 - Remove unnecessary else after guard condition
 - Replaces an unused index in a for loop or comprehension with an underscore
 - Merges together multiple nested if conditions into one
 - Replaces use of default mutable arguments in a function
 - Replaces dictionaries created with dict() with {}
 - Swap the if and else branches of an if-expression to remove a negation (easier to read)
 - Add additional web server to test/perf using aiohttp `test/perf/http_server_aiohttp.py`
 - Fix the test/perf sanic webserver
 - Consolidates multiple comparisons into a single comparison
 - Update pre-commit config to use latest versions


mite/context.py
```diff
@@ -50,9 +50,7 @@ def config(self):
 
     @property
     def should_stop(self):
-        if self._should_stop_func is not None:
-            return self._should_stop_func()
-        return False
+        return False if self._should_stop_func is None else self._should_stop_func()
```
aecay, line 53: `self._should_stop_func and self._should_stop_func() or False` might be an alternative

mite/runner.py
```diff
@@ -204,8 +204,7 @@ async def _execute(self, context, scenario_id, scenario_data_id, journey_spec, a
                 else:
                     await journey(context, *args)
         except Exception as e:
-            if not getattr(e, "handled", False):
-                if self._debug:
-                    breakpoint()
-                    raise
+            if not getattr(e, "handled", False) and self._debug:
```
aecay, line 207: you could reorder: `if self._debug and getattr(...)`.  this has the advantage of not needing to make a getattr call if debug is false.

mite/web/prometheus.py
```diff
@@ -70,23 +69,23 @@ def update(self, message):
...
                 labels = format_dict(dict(zip(self.labels, key)))
-                for bin_label, bin_count in zip(self.bins, bin_counts):
-                    lines.append(
-                        '%s_bucket{%s,le="%.6f"} %d'
-                        % (self.name, labels, bin_label, bin_count)
-                    )
+
+                lines.extend(
+                    '%s_bucket{%s,le="%.6f"} %d'
+                    % (self.name, labels, bin_label, bin_count)
```
aecay, line 82: no fstring? :wink:

aecay:
these are generally good changes.  a few of them are stylistically charged.  i don't feel like exercising any kind of direction over the stylistic evolution of the mite code base, but i thought it would be good to point out which i see as the ones that have debatable value.

> Replace Interpolation With Fstring

this is a clear win when it's replacing multiple substitutions, but when it replaces something like `variable + "constant"` the readability suffers.  `x+y` is very clear as to what it accomplishes (adds two strings together) but `f"..."` requires visually parsing the whole string to see how many interpolations there are.

> Returning a result directly instead of assigning a variable then returning that

sometimes these are debugging leftovers.  if you do `return some(expression)`, then there's no way to stick in a cheeky `print(the_value_of_some_expression)` or just `breakpoint()` into the code.  i don't know if that's a good reason to keep the extra assignment around, but i imagine at least in some fraction of the cases, someone is eventually going to have to undo this transformation to debug something.

> Merge assignment followed by conditional check using a named expression. Cleaner code

asserting that this is cleaner code is debatable (and hotly debated).  i don't think there are too many objective grounds to have the discussion on.  but it would be much clearer if you proposed a change in the project's style to prefer the walrus.  insofar as there was a style before, it was to disprefer the walrus, mostly for reasons of conservatism (the code base predates the walrus).  if there's a clear style (bonus points: enforced by a linter), then everyone's on the same page about what to write and there's less risk of future drift (requiring periodic PRs like this one to re-baseline everything).

a few more comments left inline.

aecay:
> could be personal preference

It definitely is!  The case that bothers me is

```
f"{my_variable} plus some other long stuff that is very long"

vs

my_variable + " plus some other long stuff that is very long"
```

The second is clearer (at least to me) that there's one variable and one constant being combined; the first one needs a bit more study to determine what is going on (especially if "long stuff" has some non-alphabetic characters in it that you have to visually parse to make sure they are not `{}`)  (Also: if your text editor syntax highlights f strings in a way that distinguishes `{...}`, then it looks more distinct...but it also needs to be readable e.g. in git diffs where there's no syntax highlighting.)  But it's definitely not a reason to block your changes, just something to consider.

> If you feel strongly about keeping the "assign then return" logic

Nope, I think it's fine to go ahead with your change.  It might be something to keep an eye on as to whether these creep back into the code over time.

> we could only support 3.9+

We used to have a policy to support N-1 versions of python (so, the current release = 3.10 and one previous = 3.9), so this would be ok.  Syntactic evolution of python (impressionistically) has sped up recently though, so I don't know if that is still a good rule.  IMO `|=` instead of dictionary update is a less thorny change than the walrus, so if we were happy to commit to 3.9 then I'd say go ahead.

As for the walrus itself, I'm not fully sold either (if you couldn't tell) but it does seem to be where the ecosystem is going.  So we probably ought to try it out now while we [=you] are making these changes.  We [=you again :sweat_smile:] can always rip it all out again if it causes too many problems...

tldr I'd personally be ready to press :heavy_check_mark: onthis when it's ready

Dockerfile.perftest
```diff
@@ -1,4 +1,4 @@
-FROM python:3.9
+FROM python:10
```
aecay, line 1:
```suggestion
FROM python:3.10
```

aecay (approved):
one tiny change seems needed but overall this lgtm

Dockerfile.perftest
ryanlinnit-sky: Good spot! Fixed 😃

PR #241, opened 2022-07-12
Allow `mite cat` to cat bytes (#241) 🐱

ryanlinnit-sky:
I'm not sure if it's related to using the new acurl, but.. the http method was getting sent with the type bytes, causing `mite cat` to fail, e.g.

```
$ mite cat collector_data/current
{"start_time": 1657611774, "effective_url": "http://localhost:9898/", "response_code": 200, "dns_time": 6e-06, "connect_time": 6e-06, "tls_time": 0.0, "transfer_start_time": 1.4e-05, "first_byte_time": 0.000248, "total_time": 0.000252, "primary_ip": "127.0.0.1", "method": Traceback (most recent call last):
  File "/Users/rli14/mite/mite_env/bin/mite", line 33, in <module>
    sys.exit(load_entry_point('mite', 'console_scripts', 'mite')())
  File "/Users/rli14/mite/mite/__main__.py", line 352, in main
    cat(opts)
  File "/Users/rli14/mite/mite/cli/cat.py", line 24, in cat
    json.dump(row, sys.stdout)
  ...
TypeError: Object of type bytes is not JSON serializable
```

This PR fixes `mite cat` to allow it to handle byte values.

### DIFF
--- a/mite_http/__init__.py
+++ b/mite_http/__init__.py
@@ -81,7 +81,7 @@ class SessionPool:
                 first_byte_time=r.starttransfer_time,
                 total_time=r.total_time,
                 primary_ip=r.primary_ip,
-                method=r.request.method,
+                method=r.request.method.decode("utf-8"),
                 **session_wrapper.additional_metrics,
             )
 
--- a/mite_http/__init__.py
+++ b/mite_http/__init__.py
@@ -81,7 +81,7 @@ class SessionPool:
                 first_byte_time=r.starttransfer_time,
                 total_time=r.total_time,
                 primary_ip=r.primary_ip,
-                method=r.request.method.decode("utf-8"),
+                method=r.request.method,
                 **session_wrapper.additional_metrics,
             )
 
--- a/mite/cli/cat.py
+++ b/mite/cli/cat.py
@@ -21,7 +21,7 @@ def prettify_timestamps(d):
 
 
 def cat(opts):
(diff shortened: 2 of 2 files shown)


Sarah-J-S (approved):

mite_http/__init__.py
```diff
@@ -81,7 +81,7 @@ def response_callback(r):
                 first_byte_time=r.starttransfer_time,
                 total_time=r.total_time,
                 primary_ip=r.primary_ip,
-                method=r.request.method,
+                method=r.request.method.decode("utf-8"),
```
aecay, line 84: i avoided doing this for efficiency reasons.  i don't have benchmarks to show how much it matters, but the theory was that avoiding extra work was a good thing.  if `mite cat` is the only use case that cares, then i'd prefer to fix it there.  (i seem to remember that some prometheus metrics got uglier too, as in `method="b'GET'"` instead of just `method="GET"`...idk if that was ever fixed but it would be presumably trivial to do so in the prometheus metrics code.  the point is not to do it on the hot path.)

ryanlinnit-sky: I did some benchmarking and it's pretty negligible, but still impacting. I think I agree, if it only breaks `mite cat` it's not worth it and this PR will allow `mite cat` to output byte encoding values anyway.

aecay (approved):
lgtm

PR #243, opened 2022-08-11
Moving CI from Jenkins to CircleCI

ryanlinnit-sky:
Moving CI from Jenkins to CircleCI

**Jobs**

- check-acurl-version
	- Checks if acurl has changed, if so check the version has updated in acurl/setup.cfg
- build-and-test-3{8,9,10}
	- Installs all dependancies
	- Runs pre-commit checks
	- Run `tox -e py3{8,9,10}`
- tag
	- Only runs on master
	- run `cd-scripts/cdRelease.sh` script to increase version and tag on github
- linux-wheels
	- Only runs on master
	- Build mite wheels and if acurl has changed, build acurl wheels and source package
	- Upload them to PyPi
- osx-wheels
	- Only runs on master
	- If acurl has changed, build acurl wheels (x86_64,ARM64) and upload them to PyPi

**Notes**
Both mite and acurl originally used `use_scm_version` for their version, which I don’t think we want if during the pipeline we automatically bump the version and create a tag in github for mite. This would mean any small change in mite causes a version bump in acurl. The alternate and what this PR does, is allow acurl to use the version set in `acurl/setup.cfg` but any changes done in acurl, must be accompanied with a manual update of the version.

Before this PR is merged, the the SSH key for the tag job needs to be updated to a key that has write access to the mite repo. A CircleCI project needs to be setup for the mite repo and the `PYPI_PASSWORD` will need setting.


acurl/setup.py
```diff
@@ -28,6 +28,5 @@
             "warn.unused_result": True,
         },
     ),
-    setup_requires=["setuptools_scm", "cython"],
```
jb098, line 31: What was setuptools_scm giving us before?

cd-scripts/cdRelease.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash
```
aecay, line 1: could we write this in python instead?  (or any sane language, although python is the obvious choice in this context...but my main motivation here is bash-avoidance rather than python-suprematism)

.circleci/config.yml
```diff
@@ -0,0 +1,175 @@
...
+          pkg-manager: pip
+      - python/install-packages:
+          pip-dependency-file: dev-requirements.txt
+          pkg-manager: pip
+      - run:
+          name: pre-commit checks
+          command: pre-commit run --origin HEAD --source origin/master || git diff
+      - run:
+          name: tox
+          command: tox -e py310
```
aecay, line 90: do we care about also running the tests under other python versions?  i think we've caught the occasional compatibility error that way.  that's supposed to be what tox supports.  admittedly the old test pipeline could not do this, but that's due to the way it was shackled to a jenkins image that made it very difficult to install multiple pythons.  now we could either use a multi-python image to run the tests in (if such a thing is available), or else have multiple test jobs for 3.9 and 3.10.

acurl/setup.py
ryanlinnit-sky: `0.1.dev231+g6f2b85e` as we don't currently have/set a version in git, it looks ugly like this.

If we create a git tag, we can use this and it looks nice (e.g. 1.0.0), which is what I suggest we do for mite itself.

Also PyPi doesn't accept this as a version and you can't upload packages with the version looking like this.

cd-scripts/cdRelease.sh
ryanlinnit-sky: Good point, I'll rewrite this in Python. (Although I don't hate bash 😊)

ryanlinnit-sky: Done 🥳

aecay (approved):
these changes lgtm!  great work :slightly_smiling_face:

.circleci/config.yml
ryanlinnit-sky: I can look into it. I don't think there's a multi-python image, but I'll check if there's a good/easy method for setting one up (i.e. using a python 3.10 image and install 3.8 & 3.9 along side). or go the easy route and like you suggest, multiple jobs.

ryanlinnit-sky: I tried using pyenv, but couldn't get it working reliably. I've gone for multiple jobs instead, but I'll keep it on my list of improvements to make, as ideally it should be one job.

DavAnnunz (approved):

PR #251, opened 2022-11-09
Build aarch64 linux wheels for acurl

ryanlinnit-sky:
#### What is the change?

Currently the CircleCi job only build x86_64 linux wheels and x86_64/arm64 mac wheels. This change will add the build of aarch64 wheels for linux.

The use case for this is using a linux docker image on an M1 mac (without having to use a x86_64 image and the performance degradation that involves). Or a native linux os on aarch64 architecture.

- [ ] Major
- [ ] Minor
- [ ] Patch


aecay (approved):
nice!

PR #278, opened 2024-08-19
Stddev benchmark

Gopikarthicksky:
#### What is the change?

Add stddev and +/- std dev to benchmark report. 

Added new flags `--standard-deviation-response-time-threshold` and `--standard-deviation-req-sec-threshold` 

Updated mean to use statistics library and to calculate values from overall response time instead of calculating from resp_time_mean_store which has the collection of mean of response time from each instance. 
Same logic is applied for req per sec. Updated the format of report at the end. Now the report looks like this 

Screenshot 2024-09-04 at 2 15 54 PM

- [ ] Major
- [x] Minor
- [ ] Patch

### DIFF
--- a/mite/logoutput.py
+++ b/mite/logoutput.py
@@ -2,7 +2,6 @@ import logging
 import math
 import time
 from collections import defaultdict
-import numpy as np
 import statistics
 
 
@@ -58,7 +57,6 @@ class GenericStatsOutput:
         self._resp_time_max = 0
         self._resp_time_mean = 0
         self._resp_time_store = []
-        self._resp_time_mean_store = []
         self._req_sec_store = []
         self._req_sec_min = 0
         self._req_sec_max = 0
@@ -111,8 +109,6 @@ class GenericStatsOutput:
                 + f"max:{self._pct(100)}",
             )
         self._start_t = t
-        # breakpoint()
-        # self._resp_time_store.extend(self._resp_time_recent)
         del self._resp_time_recent[:]
         self._req_recent = 0
         self._error_recent = 0
@@ -129,15 +125,11 @@ class GenericStatsOutput:
             if self._resp_time_recent:
                 self._resp_time_store.extend(self._resp_time_recent)
                 if self._resp_time_min == 0:
-                    self._resp_time_min = self.min_resp_time
+                    self._resp_time_min = self.min_resp_time_recent
                 else:
-                    self._resp_time_min = min(self.min_resp_time, self._resp_time_min)
+                    self._resp_time_min = min(self.min_resp_time_recent, self._resp_time_min)
 
                 self._resp_time_max = max(self.max_resp_time_recent, self._resp_time_max)
-
-                self._resp_time_mean_store.append(
-                    sum(self._resp_time_recent) / len(self._resp_time_recent)
-                )
 
             if self._req_recent:
                 req_sec = self._req_recent / (t - self._start_t)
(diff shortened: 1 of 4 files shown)


mite/logoutput.py
```diff
@@ -173,7 +192,66 @@ def min_resp_time(self):
...
+    
+    @property
+    def req_sec_standard_deviation(self):
+        if self._req_total == 0:
+            return 0
+        variance = []
+        for req_sec in self._req_sec_store:
+            variance.append(math.sqrt((req_sec-self.req_sec_mean)**2))
+        std_deviation = sum(variance) / len(self._req_sec_store)
+        return std_deviation
```
aecay, line 221: should we use the python standard library here?  https://docs.python.org/3/library/statistics.html#statistics.pstdev

numpy is a beefy dependency which sometimes has build complications (e.g. on alpine -- i personally hate alpine but sometimes we need to use it).  open coded implementations of numerical algorithms like this one can sometimes have numerical stability issues.  i'm not knowledgable enough to say whether that affects this code, but i always opt for using a library when available because that means it's Someone Else's Problem™ to tune the algorithm.  thankfully python makes this available in the standard library, so we don't need to add a dependency to make use of it.

mite/logoutput.py
```diff
@@ -173,7 +192,66 @@ def min_resp_time(self):
...
+    #     # print(f"\nresp_time_store is {self._resp_time_store} \n")
+    #     # print(f"len of resp time store is {len(self._resp_time_store)}\n")
+    #     # del self._resp_time_store[:]
+    #     return stats_percentage
+    
+    @property
+    def resp_time_within_standard_deviation(self):
+        if not self._resp_time_store:
+            return 0
+        return np.mean(((self.resp_time_array <= (self.mean_resp_time + self.resp_time_standard_deviation)) & (self.resp_time_array >= (self.mean_resp_time - self.resp_time_standard_deviation)))* 100)
```
aecay, line 241: i have no idea what this algorithm is doing...can you explain what it's trying to calculate?  you seem to be mixing booleans with bitwise operators, which points to very subtle things going on....

mite/logoutput.py
Gopikarthicksky: I used numpy since it is faster but come to think of it, this is supposed to do operations on smaller datasets so I guess statistics library should do the job. I'll change this

mite/logoutput.py
Gopikarthicksky: It is going to return the percentage of values that lie between the upper and lower limits of deviation. 

(self.resp_time_array <= (self.mean_resp_time + self.resp_time_standard_deviation)) & (self.resp_time_array >= (self.mean_resp_time - self.resp_time_standard_deviation) creates a boolean array where each element is True if the condition is met and False otherwise. np.mean should return the fraction of True values

aecay: I see.  Could we use something like `abs(x - self.mean_resp_time) < self.std_dev`?  It may be a matter of taste (so feel free to tell me you disagree), but at least to me that's a clearer way of writing the algorithm than `mean-stddev < x < mean+stddev`.

Whatever way we decide to code it, I think a comment explaining what the calculation does would be really helpful too.

Gopikarthicksky: This is actually good and can be easily understood. I'll make changes

aecay (approved):
lgtm, thanks!

jb098 (approved):

PR #280, opened 2024-08-20
Acurl use Cython 3.0.8

ryanlinnit-sky:
#### What is the change?

Add compatibility for Cython version greater than 3.

This is only a version increment for acurl, done in the `acurl/setup.cfg` file. Not mite itself.

- [ ] Major
- [ ] Minor
- [ ] Patch


acurl/src/acurl.pyx
```diff
@@ -21,7 +21,7 @@ class AcurlError(Exception):
 # with the python asyncio library we need to call back into python.  So these
 # functions do take the GIL, as indicated by `with gil` in their signatures.
 # The performance impact of this is yet tbd.
-cdef int handle_socket(CURL *easy, curl_socket_t sock, int action, void *userp, void *socketp) with gil:
+cdef int handle_socket(CURL *easy, curl_socket_t sock, int action, void *userp, void *socketp) noexcept with gil:
```
aecay, line 24: I think it might help to have some commentary about why we think noexcept is correct for these functions.  For instance, this one has code that explicitly raises an exception (although it's expected never to be hit...)  The other functions that call python code might raise in subtler ways.

I understand why we are doing this (I think) -- because we need to use these fns as C callbacks (in a context where exceptions won't get handled anyway).  But we can help spell that out for future readers of this code, especially those who might be confused seeing `raise` in a `noexcept` function, and not immediately be able to call to mind the relevant facts about the usage of callbacks etc.

WDYT?

acurl/src/session.pyx
```diff
@@ -32,7 +32,7 @@ cdef BufferNode* alloc_buffer_node(size_t size, char *data):
     node.next = NULL
     return node
 
-cdef size_t header_callback(char *ptr, size_t size, size_t nmemb, void *userdata):
+cdef size_t header_callback(char *ptr, size_t size, size_t nmemb, void *userdata) noexcept:
```
aecay, line 35: OTOH, this function (and the one below it in this file) really can't raise, and so are valid `noexcept` candidates with no explanation needed.

setup.cfg
```diff
@@ -28,6 +28,7 @@ install_requires =
     selenium<4
     uvloop
     websockets
+    setuptools
```
aecay, line 31: alphabetical order pls 🙏

acurl/setup.cfg
```diff
@@ -8,16 +8,16 @@ classifiers =
...
     Programming Language :: Python
     Programming Language :: Python :: 3
-    Programming Language :: Python :: 3.9
     Programming Language :: Python :: 3.10
     Programming Language :: Python :: 3.11
+    Programming Language :: Python :: 3.12
 author = Aaron Ecay
 author_email = aaron.ecay@sky.uk
 maintainer = Sky Identity NFT team
 maintainer_email = matthew.ellis@sky.uk
```
aecay, line 17: I guess we can fix this up while we're here -- do you want to be maintainer of acurl? 😅

acurl/pyproject.toml
```diff
@@ -1,2 +1,2 @@
 [build-system]
-requires = ["setuptools", "wheel", "Cython<3.0"]
+requires = ["setuptools", "wheel", "Cython==3.0.8"]
```
aecay, line 2: can we pin to a version range?  pinning to an exact version means we won't pick up future bugfixes...

setup.cfg
ryanlinnit-sky: Done

acurl/pyproject.toml
ryanlinnit-sky: Pinned to `<=3.0.8,>4.0.0`

acurl/setup.cfg
ryanlinnit-sky: Do you want leaving as the "author" and just the maintainer updated? Jordan has a branch with an [updated email address](https://github.com/sky-uk/mite/blob/f46729dc3c1439ef8c59395ef8f39b82858c37f8/acurl/setup.cfg#L17).

acurl/src/acurl.pyx
aecay: We could panic rather than raising an exception, I suppose...the only way we'd hit that branch of code is if there was a breaking change to the curl API, or there was some kind of data corruption.  In either case I think we don't want it to be silent.  (I guess til now the exception would have been ignored, so it was a false sense of safety).

> is the second or action == CURL_POLL_INOUT in this function is redundant

No, since if the action is `INOUT` we need to add both the reader and the writer to the event loop.  There are two (actually three) separate `if`s, it's not an `elif`.  I guess we could refactor to use one of the fancy new `match` statements (but that would lead to code duplication, since we'd need the `add_reader` and `add_writer` calls twice, once in the separate `IN` or `OUT` case and once in the combined `INOUT` case).

setup.cfg
```diff
@@ -26,7 +26,8 @@ install_requires =
     nanomsg
     pyzmq
     selenium<4
-    uvloop
+    setuptools
+    uvloop==0.21.0b1
```
aecay, line 30: hmm, do we know why we are pinning to a specific version, especially a beta one?  what are the conditions in the future under which we could remove the pin?

test-requirements.txt
```diff
@@ -14,4 +14,4 @@ pytest-httpbin
 pytest-httpserver
 sanic
 werkzeug==3.0.3
-uvloop==0.19.0
+uvloop==0.21.0b1
```
aecay, line 17: eek -- the test should not be pinned to a different version than the main code, so it's good to change this (tho no pinning would be ideal).

it's been a while since i looked at this code, but IIRC, the tests install everything from setup.cfg as well as everything from the test-requirements.txt.  so given that uvloop is already required by `setup.cfg`, we could delete it from here and avoid any future iterations of the different-pinning problem.  WDYT?

aecay (commented):
overall the change looks great.  just one question (well, a series of them...) about some dependency pinning details.

test-requirements.txt
ryanlinnit-sky: I was getting errors due to it not the version of uvloop required for Cython 3. ([This commit)](https://github.com/MagicStack/uvloop/commit/3fba9fab1e8d66c333cf3946d79e29b5bad8fa73). Which is why I pinned it to version `0.21.0b1`. I'll put this PR back in drafts whilst I fix them etc.

aecay: ahhh, uvloop doesn't have a stable release for cython3 yet.  that explains it.  do we want to wait for upstream to cut a stable release?  that should avoid us needing to do any pinning.  if we don't want to wait, we can just stick a comment in explaining the situation, and Someone™ can remove the comment and the pin in the future....

PR #290, opened 2025-02-20
Fix data creation example in the docs

sky-jm:
#### What is the change?
Fixed minimum examples in `journey.rst` that were not consistent or wrong and therefore didn't run without edits. 

- [ ] Major
- [ ] Minor
- [x] Patch


DavAnnunz (approved):
LGTM, thank you!

jb098 (approved):
Thanks for picking up and always looking for feedback. Let us know if you need any support.

PR #312, opened 2025-11-13
Fix 92 stream error

Kes-trel:

- [ ] Major
- [ ] Minor
- [x] Patch


Kes-trel:
No longer required, was fixed part of acurl version update PR

PR #325, opened 2025-12-11
poc for ui

Gopikarthicksky:

- [ ] Major
- [ ] Minor
- [x] Patch


jb098:
A more general question but should we include this code in mite or have it in a separate repo? Maybe a question to the whole team.

PR #329, opened 2026-08-31
influxdb exporter

Gopikarthicksky:

- [ ] Major
- [x] Minor
- [ ] Patch

### DIFF
--- a/mite/cli/influxdb.py
+++ b/mite/cli/influxdb.py
@@ -22,7 +22,10 @@ def influxdb_init(opts):
 
     all_stats = []
     for ep in importlib.metadata.entry_points(group="mite_stats"):
-        all_stats.extend(ep.load())
+        try:
+            all_stats.extend(ep.load())
+        except Exception as e:
+            logger.warning(f"Failed to load stats from entry point {ep.name}: {e}")
 
     points = []
     tns = time_ns()
--- a/mite/cli/influxdb.py
+++ b/mite/cli/influxdb.py
@@ -1,0 +1,77 @@
+import importlib.metadata
+import logging
+import os
+from time import time_ns
+
+from influxdb_client_3 import InfluxDBClient3, Point
+
+logger = logging.getLogger(__name__)
+
+
+def influxdb_init(opts):
+    host = os.getenv("INFLUXDB_HOST")
+    database = os.getenv("INFLUXDB_DATABASE")
+    token = os.getenv("INFLUXDB_TOKEN")
+
+    if not token or not host or not database:
+        raise ValueError("INFLUXDB_HOST, INFLUXDB_DATABASE, and INFLUXDB_TOKEN variables are empty")
+
+    logger.info(f"Connecting to {host}, database: {database}")
+
+    client = InfluxDBClient3(host=host, token=token, database=database)
+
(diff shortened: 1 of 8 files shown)


mite/web/influxdb.py
```diff
@@ -0,0 +1,389 @@
...
+
+@dataclass
+class InfluxPoint:
+    measurement: str
+    tags: dict
+    fields: dict
+    time_ns: int
+
+
+class InfluxdbWriter:
```
jb098: Looking at this, I'm wondering if it would be cleaner to have a _InfluxBaseClass or interface and we implement that for both a v2 and v3. It would probably end up with a bit more code, but the v2 and v3 actions would be nicely encapsulated in their own classes and it may end up a bit more readable/maintainable.

mite/web/influxdb.py
```diff
@@ -0,0 +1,389 @@
...
+        delta_total = max(0, self.total_counts[k] - self._prev_total_counts[k])
+        delta_sum = max(0.0, self.sums[k] - self._prev_sums[k])
+        return delta_bin_counts, delta_total, delta_sum
+
+    def to_points(self, time_ns, touched_keys=None):
+        points = []
+        with self._lock:
+            keys_to_iterate = self.total_counts if touched_keys is None else touched_keys
+
+            for k in sorted(keys_to_iterate):
```
jb098, line 348: I'm wondering if iterating over this might be more performant if we use something like a map() to do them all at once, breaking all the logic into another function which is applied to each item in the iterable by the map()

mite/cli/influxdb.py
```diff
@@ -0,0 +1,85 @@
...
+            points.append(InfluxPoint(
+                measurement=name,
+                tags=tags,
+                fields={"value": 0.0},
+                time_ns=tns,
+            ))
+
+        elif stat_type == "Histogram":
+            fields = {}
+            for pct in [50, 75, 90, 95, 99]:
```
jb098, line 45: We often display 99.9 and 99.99 when using prometheus, though I understand that's because the bucketting behaves differently.

We could argue we should include some of that but I think we can add later if we want it.

mite/cli/test.py
```diff
@@ -75,6 +76,13 @@ def _setup_msg_processors(receiver, opts):
     extra_processors = [
         spec_import(x)(opts) for x in opts["--message-processors"].split(",") if x
     ]
+
+    if opts["--influxdb"]:
+        influxdb_count = int(opts["--influxdb"])
+        for i in range(influxdb_count):
```
jb098, line 82: Rather than having an extra line for count here, you could use enumerate()

jb098 (approved):

mite/cli/test.py
Gopikarthicksky: influxdb_count is literally used as a counter here. enumerate would just give me (1,1) (2,2) and so on. Might not be the right use case here. Lmk if your vision differs from mine

mite/cli/influxdb.py
Gopikarthicksky: you are right. because of how bucket bounds are written here having 99.9 or 99.99 would merely be a tiny noise

